PR #5, opened 2017-09-07
Add uninstall app feature + tests

CPatchane:


CPatchane:
[(3) [STO-4] ETU je peux désinstaller une application](https://trello.com/c/FJuYTaNe/537-3-sto-4-etu-je-peux-d%C3%A9sinstaller-une-application)

src/components/ApplicationModal.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+import React, { Component } from 'react'
+
+import { translate } from 'cozy-ui/react/I18n'
+import Modal, { ModalContent } from 'cozy-ui/react/Modal'
+
+import ReactMarkdownWrapper from './ReactMarkdownWrapper'
+
+class ApplicationModal extends Component {
+  uninstallApp (appSlug) {
```
gregorylegarec: Shouldn't `appSlug` be in `props` ?

src/containers/MyApplications.jsx
```diff
@@ -11,9 +11,8 @@ const mapStateToProps = (state, ownProps) => ({
 })
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
-  fetchApps: () => {
-    dispatch(fetchApps())
-  }
+  fetchApps: () => dispatch(fetchApps()),
+  uninstallApp: (appSlug) => dispatch(uninstallApp(appSlug))
```
gregorylegarec: IMHO `uninstallApp` should not have parameter as the whole component is contextualized with an `appSlug`. What do you think ?
```javascipt
uninstallApp: () => dispatch(uninstallApp(ownProps.appSlug))
```

src/ducks/index.js
```diff
@@ -1,9 +1,11 @@
 import { combineReducers } from 'redux'
 
 import { myAppsReducers } from '../ducks/myApps'
+import alerterReducer from 'cozy-ui/react/Alerter'
 
 const storeApp = combineReducers({
-  myApps: myAppsReducers
+  myApps: myAppsReducers,
+  alerts: alerterReducer
```
gregorylegarec: I think we can use 
```js
const storeApp = combineReducers({
  myApps: myAppsReducers,
  alerterReducer
})
```

src/ducks/myApps/index.js
```diff
@@ -90,6 +90,15 @@ export function fetchApps () {
 export function uninstallApp (slug) {
   return (dispatch, getState) => {
     return cozy.client.fetchJSON('DELETE', `/apps/${slug}`)
+    .then(() => {
+      return dispatch({
+        type: 'SEND_LOG_SUCCESS',
+        alert: {
+          message: 'app_modal.uninstall.alerts.success',
```
gregorylegarec: To me it's weird to send an `alert`, an UI related information, from this.

IMHO we should instead have a part of the state reacting to a list of actions, instead of passing those kind of parameters from here.

gregorylegarec: Or maybe `alert` is too much here, we only need `message`.

src/locales/en.json
```diff
@@ -22,7 +22,10 @@
       "title": "Uninstall this application",
       "description": "This application is bound to be removed from your Cozy %{cozyName}. Are you sure to __remove this application definitively__?",
       "cancel": "Cancel",
-      "uninstall": "Uninstall"
+      "uninstall": "Uninstall",
+      "alerts": {
```
gregorylegarec: Same here, having `alerts` here is too much behavior related. What if towmorrow we do not use alert anymore ?
I think we should only have something like:
```js
"app_modal": {
  "uninstall": {
    "success": {
      "message": "The application has correctly been uninstalled."
     }
   }
}
```
By the way, if I understand well key structure, we have somewhere a key `app_modal.uninstall.uninstall` ?

src/ducks/myApps/index.js
```diff
@@ -94,7 +94,7 @@ export function uninstallApp (slug) {
       return dispatch({
         type: 'SEND_LOG_SUCCESS',
         alert: {
-          message: 'app_modal.uninstall.alerts.success',
+          message: 'app_modal.uninstall.message.success',
```
gregorylegarec: Better 😄

src/locales/en.json
```diff
@@ -23,8 +23,9 @@
       "description": "This application is bound to be removed from your Cozy %{cozyName}. Are you sure to __remove this application definitively__?",
       "cancel": "Cancel",
       "uninstall": "Uninstall",
-      "alerts": {
-        "success": "The application has correctly been uninstalled."
+      "message": {
```
gregorylegarec: ❤️

src/ducks/myApps/Container.jsx
```diff
@@ -12,7 +12,10 @@ const mapStateToProps = (state, ownProps) => ({
 
 const mapDispatchToProps = (dispatch, ownProps) => ({
   fetchApps: () => dispatch(fetchApps()),
-  uninstallApp: (appSlug) => dispatch(uninstallApp(appSlug))
+  uninstallApp: (appSlug) => {
+    return dispatch(uninstallApp(appSlug))
+    .then(() => dispatch(fetchApps()))
```
gregorylegarec: Do we really need to fetch ? `uninstallApp` should remove the app from the state, shouldn't it ?

src/ducks/components/SmallAppItem.jsx
```diff
@@ -1,20 +1,23 @@
...
+  const appVersion = version && (version.match(/^(.*)-.*$/) ? version.match(/^(.*)-.*$/)[1] : version)
+  return (
+    <div className='sto-small-app-item' onClick={onClick} >
+      <img src={icon} alt={`${slug}-icon`} width='64' height='64' className='sto-small-app-item-icon' />
+      <div className='sto-small-app-item-desc'>
+        <h4 className='sto-small-app-item-title'>
+          {editor && `${editor} `}{name}
+        </h4>
+        <p className='sto-small-app-item-detail'>
+          {developer.name === 'Cozy' ? 'Cozy Cloud Inc.' : developer.name}
```
gregorylegarec: I think we should never hardcode UI text in our code. We should always use translations instead.

src/components/ApplicationModal.jsx
CPatchane: That could be, you're right 👍

src/containers/MyApplications.jsx
CPatchane: In fact, the `appSlug` is grabbed from the router, so it's directly available in the props like that. It's why I use it as parameter here.

src/ducks/components/SmallAppItem.jsx
CPatchane: I agree, but it's an exception I think, we will always display it like this, won't we?

src/locales/en.json
CPatchane: Yep, I didn't notice that but it's the "Uninstall" button in the modal.

src/ducks/components/SmallAppItem.jsx
gregorylegarec: By the way why do we have to do this ? IMHO this is an information that should be in the manifest whatever happen, don't you think ? Even if we need to add a `displayName` property to `developer`.

CPatchane: I agree, I hope that will be a temporary code here. But by waiting to be in our manifests as developer name, it does the job :/

gregorylegarec:
Thanks @CPatchane ! 🎃

PR #9, opened 2017-09-14
Install app feature 

CPatchane:


m4dz:
@CPatchane still issues with SVG assets during watches, or it's ok?

CPatchane:
@m4dz It's still here, or maybe it's just on my side. If you could try on your side that would be great :)

src/ducks/myApps/Container.jsx
```diff
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux'
 
-import { fetchApps, uninstallApp } from './index'
+import { fetchMyApps, uninstallApp } from './index'
```
gregorylegarec: I hate the `My` prefix for variable or Classes. I think it is too confusing. What about `InstalledApps` here ?

src/ducks/index.js
```diff
@@ -1,10 +1,10 @@
 import { combineReducers } from 'redux'
 
-import { myAppsReducers } from '../ducks/myApps'
+import { appsReducers } from '../ducks/myApps'
 import alerterReducer from 'cozy-ui/react/Alerter'
 
 const storeApp = combineReducers({
-  myApps: myAppsReducers,
+  apps: appsReducers,
   alerts: alerterReducer
```
gregorylegarec: What about:
```js
import { combineReducers } from 'redux'

import apps from './apps'
import currentAppVersion from './apps/currentAppVersion'
import alerts from 'cozy-ui/react/Alerter'

const storeApp = combineReducers({
  apps
  currentAppVersion,
  alerts
})

export default storeApp
```
Ducks are supposed to export a default function named `reducer` https://github.com/erikras/ducks-modular-redux

src/ducks/myApps/index.js
```diff
@@ -8,29 +8,37 @@ import {
   NotUninstallableAppException
 } from '../../lib/exceptions'
 
-const FETCH_MY_APPS = 'FETCH_MY_APPS'
-const FETCH_MY_APPS_SUCCESS = 'FETCH_MY_APPS_SUCCESS'
-const FETCH_MY_APPS_FAILURE = 'FETCH_MY_APPS_FAILURE'
+const NOT_REMOVABLE_APPS = ['drive', 'collect']
```
gregorylegarec: Aren't we supposed to get those kind of information from the stack ? I suggest to add a comment to remember talking to them about that point.

src/ducks/myApps/index.js
```diff
@@ -39,60 +47,118 @@ const isFetching = (state = false, action) => {
...
+      return URL.createObjectURL(icon)
+    } catch (e) {
+      return ''
+    }
   }
 }
 
-const NOT_REMOVABLE_APPS = ['drive', 'collect']
-const NOT_DISPLAYED_APPS = ['settings', 'store', 'onboarding']
+function _consolidateApps (stateApps, newAppsInfos) {
```
gregorylegarec: I see your point of using an `_` prefix here, for indicating that this method is private, bet if I am right it is not a convention we are using in our other apps. So to be coherent, I suggest to remove the `_` what do you think ?

src/ducks/myApps/components/Discover.jsx
```diff
@@ -22,7 +32,12 @@ export class Discover extends Component {
           </h3>
           <div className='sto-discover-get-started-list'>
             {!isFetching && apps.map(a => {
```
gregorylegarec: Could we use an  `app` variable here ? I hate one-letter variables, even in simple cases like this one. If I look only line 42 for example I read only `slug={a.slug}` and I do not understant what we are dealing with. What do think ?

gregorylegarec:
Ha ha ha ok all my comments are outdated, so just "LGTM"

src/ducks/myApps/index.js
CPatchane: It's not only because it's private, I use it also to say explicitly that these methods are not supposed to be exported and used externally like that. If this case happens, that parts should be moved in a external library/file.

src/ducks/myApps/Container.jsx
CPatchane: Completely agree 👍

src/ducks/myApps/index.js
CPatchane: I don't think so. These apps can't be removed because they are a necessary part of the Cozy platform itself and not due to a specific context. But we could discuss about that.

CPatchane:
[(5) [STO-5] Je peux installer une application](https://trello.com/c/NbncDTxQ/594-5-sto-5-je-peux-installer-une-application)

PR #10, opened 2017-09-17
Improvements + add hidden installer view + better tests

CPatchane:


CPatchane:
[(5) [STO-5] Je peux installer une application](https://trello.com/c/NbncDTxQ/594-5-sto-5-je-peux-installer-une-application)

src/ducks/apps/components/Discover.jsx
```diff
@@ -32,12 +32,8 @@ export class Discover extends Component {
...
           <div className='sto-discover-get-started-list'>
             {!isFetching && apps.map(app => {
-              // don't display app installed but not in registry
-              if (!app.isInRegistry) return null
-              let version = app.version
-              if (app.versions && Array.isArray(app.versions.stable)) {
-                version = (app.versions.stable && app.versions.stable[app.versions.stable.length - 1])
-              }
+              const stableVers = app.versions.stable
+              const version = stableVers[stableVers.length - 1]
```
gregorylegarec:
```js
const version = stableVers.pop()
```

src/ducks/apps/index.js
```diff
@@ -238,10 +238,11 @@ export function uninstallApp (slug) {
   }
 }
 
-export function installApp (slug, source) {
+export function installApp (slug, source, isUpdate = false) {
   return (dispatch, getState) => {
     dispatch({type: INSTALL_APP})
-    return cozy.client.fetchJSON('POST', `/apps/${slug}?Source=${source}`)
+    const verb = isUpdate ? 'PUT' : 'POST'
+    return cozy.client.fetchJSON(verb, `/apps/${slug}?Source=${encodeURIComponent(source)}`)
```
gregorylegarec: I think this part should at least be encapsulated into an `app.js` library, it not in CozyClient. To be also usable without redux. What do you think ?

src/ducks/apps/index.js
```diff
@@ -238,10 +238,11 @@ export function uninstallApp (slug) {
   }
 }
 
-export function installApp (slug, source) {
+export function installApp (slug, source, isUpdate = false) {
```
gregorylegarec: Should n't be more logical to have a method called `updateApp`, instead of passing a boolean parameter to a method name `installApp` ?

src/ducks/apps/Containers.jsx
```diff
@@ -1,9 +1,10 @@
 import { connect } from 'react-redux'
 
-import { fetchApps, fetchInstalledApps, uninstallApp, getInstalledApps, getRegistryApps, installAppFromRegistry } from './index'
+import { fetchApps, fetchInstalledApps, uninstallApp, getInstalledApps, getRegistryApps, installAppFromRegistry, installApp } from './index'
 
 import MyApplicationsComponent from './components/MyApplications'
 import DiscoverComponent from './components/Discover'
+import HiddenInstallerViewComponent from './components/HiddenInstallerView'
```
gregorylegarec: I don't get the role of this component. Why is it hidden ?
I feel that `View` is redundant with `Component`. What do you think ?

src/ducks/apps/components/Discover.jsx
CPatchane: Since `stableVers` is just a reference to `app.versions.stable`, I am afraid that will mute the `app.versions.stable` array here and that could be error prone :/

src/ducks/apps/Containers.jsx
CPatchane: It's the previous hidden view of cozy-settings used to install any apps using a form.
I know about the `Component` naming part but it's just for this file since it will be exported as `HiddenInstallerView` as a connected component just below.

src/ducks/apps/index.js
CPatchane: Since the update is just a change of `HTTP verb` I am really not sure about using another function for that. I would prefer to rename this method to `installOrUpdateApp`. What do you think?

gregorylegarec: We could use 3 functions, but the main one should be name `installApp`, so let's go with the actual solution 👍

src/ducks/apps/Containers.jsx
gregorylegarec: Ah ok I see. I surely miss something, I did know that this view should come back. Is it because the stack registry is not ready yet ?

CPatchane: For that and also because we could need to install an app not yet in the registry and the store won't allow it.

src/ducks/apps/index.js
CPatchane: I see your point, I think I'll have to do it in a second time but not in this PR.

gregorylegarec:
Thanks @CPatchane ! 🍰

PR #19, opened 2017-09-27
Improvements

CPatchane:
* Handle app icon from registry
* Fix bug about missing spinner in install button

### DIFF
--- a/src/ducks/apps/index.js
+++ b/src/ducks/apps/index.js
@@ -183,7 +183,7 @@ export function fetchRegistryApps (lang = 'en') {
       return Promise.all(apps.map(app => {
         const appName = (app.name && (app.name[lang] || app.name.en)) || app.slug
         const appDesc = (app.description && (app.description[lang] || app.description.en)) || ''
-        return _getIcon(app.logo_url)
+        return _getIcon(`/registry/${app.slug}/icon`)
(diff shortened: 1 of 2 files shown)


src/ducks/apps/index.js
```diff
@@ -183,7 +183,7 @@ export function fetchRegistryApps (lang = 'en') {
       return Promise.all(apps.map(app => {
         const appName = (app.name && (app.name[lang] || app.name.en)) || app.slug
         const appDesc = (app.description && (app.description[lang] || app.description.en)) || ''
-        return _getIcon(app.logo_url)
+        return _getIcon(`/registry/${app.slug}/icon`)
```
gregorylegarec: Does it mean that we are still relying on a local icon ?

gregorylegarec:
Thanks @CPatchane ⌨️

CPatchane:
[(1/5) [STO-5] Je peux installer une application](https://trello.com/c/NbncDTxQ/594-1-5-sto-5-je-peux-installer-une-application)

src/ducks/apps/index.js
CPatchane: Nope it's an icon provided by cozy-stack using this route. But we still have a local fallback icon in case of errors when fetching the app icon.

PR #26, opened 2017-10-10
feat: display description in discovery

Narkfr:


Narkfr:
[(2) [STO-7] J'ai accès à la description d'une application installable](https://trello.com/c/mwwcZIsH/597-2-sto-7-jai-acc%C3%A8s-%C3%A0-la-description-dune-application-installable)

src/ducks/apps/components/ApplicationRouting.jsx
```diff
@@ -1,28 +1,41 @@
...
-    const { installedApps, isFetching, parent, history } = this.props
+    const { apps, installedApps, isFetching, isInstalling, parent, history } = this.props
     return (
       <div>
         <Route path={`/${parent}/:appSlug`} render={({ match }) => {
           if (isFetching) return
-          if (installedApps.length && match.params) {
+          if (installedApps && installedApps.length && match.params) {
             const app = installedApps.find(app => app.slug === match.params.appSlug)
             if (!app) return history.push(`/${parent}`)
```
CPatchane: What's happen here if this is an application from the registry (so in `apps` only) but not installed, and there is some installedApps? By reading here, it seems to redirect to the parent. Am I wrong?
The `apps` should content all apps (registry + installedApps), so you can rely on it directly I think.

Narkfr: In fact `apps` and `installedApps` are passed by `Discover` and `MyApplications`, respectively, so we can't have both `apps` and `installedApps` in this component.

CPatchane: You're right, so what about checking the parent?

src/ducks/apps/components/ApplicationRouting.jsx
```diff
@@ -1,28 +1,34 @@
...
       <div>
         <Route path={`/${parent}/:appSlug`} render={({ match }) => {
           if (isFetching) return
-          if (installedApps.length && match.params) {
-            const app = installedApps.find(app => app.slug === match.params.appSlug)
+          if (appsArray.length && match.params) {
+            const app = appsArray.find(app => app.slug === match.params.appSlug)
             if (!app) return history.push(`/${parent}`)
-            return <ApplicationDetails app={app} parent={parent} />
+            return <ApplicationDetails app={app} parent={parent} installed={app.installed} />
```
CPatchane: Why do you need to pass `app.installed` since we already have this information in `app` already passed as props.

src/ducks/apps/components/ApplicationRouting.jsx
```diff
@@ -1,28 +1,34 @@
...
           }
         }} />
         <Route path={`/${parent}/:appSlug/manage`} render={({ match }) => {
           if (isFetching) return
-          if (installedApps.length && match.params) {
-            const app = installedApps.find(app => app.slug === match.params.appSlug)
-            if (!app) return history.push(`/${parent}`)
```
CPatchane: We also need to return to the parent here if no app found

src/ducks/apps/components/ApplicationDetails.jsx
```diff
@@ -24,23 +25,37 @@ export const ApplicationDetails = ({t, lang, app: { description, icon, installed
...
+              className='coz-btn coz-btn--danger-outline'
+            >
+              {t('app.uninstall')}
+            </Link>
+          </div>
+          : <Link
+            to={`/${parent}/${slug}/manage`}
+            className='coz-btn coz-btn--regular'
+          >
+            <svg className='sto-app-icon--button'>
```
CPatchane: According to the [`Icon` component](https://github.com/cozy/cozy-ui/tree/master/react/Icon) from cozy-ui, you could use it with a custom icon.
```js
<Icon icon={cozySmileWhiteIcone} />
```
Also there is a typo s/cozySmileWhiteIcome/cozySmileWhiteIcon

Narkfr: :+1:

CPatchane:
Thanks @Narkfr 🎉

PR #115, opened 2018-02-26
[WIP] Force go to parent view after uninstalling app

CPatchane:


CPatchane:
Wrong way, I close it.

PR #118, opened 2018-02-27
Default app icon if missing from registry

CPatchane:


src/ducks/apps/index.js
```diff
@@ -220,24 +220,26 @@ export function getFormattedRegistryApp (response, channel) {
...
-          // for the stack so we use appType here
-          type: version.type,
-          // add screensLinks property only if it exists
-          ...(screensLinks ? {screenshots: screensLinks} : {}),
-          installed: false,
-          uninstallable: true,
-          isInRegistry: true
-        }
-      )
+      return _getIcon(iconLink).then(iconData => {
```
gregorylegarec: You can switch from Object to Promise without any side effect ?

CPatchane: Since we were already in a promise here, it's wrapped in a promise, so I don't think so.

gregorylegarec:
Thanks @CPatchane 🏋️

PR #119, opened 2018-02-27
Update hidden installer to handle konnectors

CPatchane:


src/ducks/apps/components/HiddenInstallerView.jsx
```diff
@@ -22,8 +23,9 @@ class HiddenInstallerView extends Component {
   }
 
   handleSubmit () {
-    const { slug, source, isUpdate } = this.state
-    this.props.installUsingInstaller(slug, source, isUpdate)
+    const { slug, source, isUpdate, isKonnector } = this.state
+    const appType = isKonnector ? 'konnectors' : 'webapp'
```
gregorylegarec: Does it make sens to use constants for `konnectors` and `webapp` here ?  Are we using plural form ?

CPatchane: Good point, I already have them as constants indeed 👍

PR #156, opened 2018-04-04
Remove specific konnector slug handling

CPatchane:
Previously, we was adding `konnector-` before konnectors slug for the registry and remove it before installing in the cozy-stack.
Now we remove this specific behaviour.


gregorylegarec:
Thanks @CPatchane 🔩

PR #161, opened 2018-04-12
Filter apps according to URL parameters

CPatchane:


gregorylegarec:
What URL look like ? What I have in mind was a mechanism able to interally redirect from a query string to a "nice" url. Example `store.cozy.works?cat=mobile` to `store.cozy.works/#/category/mobile`.
What do you think ?

CPatchane:
Here are some possibilities:
```
store.cozy.works?type=webapp
store.cozy.works?doctype=io.cozy.accounts
store.cozy.works?tag=bills
store.cozy.works?category=bank
# with possibility to combine them 
store.cozy.works?type=webapp&doctype=io.cozy.accounts
```

Using this way, you can modify the search in the URL without rerendering all your app or calls the redux store. Just the applications list component will be update and it's instantaneous.
Using a redirection, you won't be able to change the query after the redirection and the application will  have to rerender more components. Since we don't have any dedicated view for searching or categories, I think it's not suitable yet.

gregorylegarec:
Thanks @CPatchane 👌

PR #168, opened 2018-04-16
feat: add intent for konnector installation ✨

gregorylegarec:
This PR :
* adds intent handling
* extract the `<AppInstallation />` component from `<InstallModal />` to use it in `<KonnectorInstallIntent />`

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -14,9 +14,24 @@
   },
   "default_locale": "en",
   "locales": {},
+  "intents": [
+    {
+      "action": "INSTALL",
+      "data": ["slug"],
+      "href": "/intents",
+      "type": [
+        "io.cozy.apps"
+      ]
+    }
+  ],
   "routes": {
     "/": {
       "folder": "/",
+      "index": "index.html",
+      "public": false
+    },
+    "/intents": {
+      "folder": "/intents",
       "index": "index.html",
       "public": false
     }
--- a/src/ducks/apps/components/AppInstallation.jsx
+++ b/src/ducks/apps/components/AppInstallation.jsx
@@ -1,0 +1,105 @@
+import React, { Component } from 'react'
+
+import PermissionsList from './PermissionsList'
+import ReactMarkdownWrapper from '../../components/ReactMarkdownWrapper'
+
+import { getLocalizedAppProperty } from 'ducks/apps'
+import { translate } from 'cozy-ui/react/I18n'
+
+class AppInstallation extends Component {
+  installApp () {
+    this.setState({ error: null })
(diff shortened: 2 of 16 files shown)


gregorylegarec:
[(5) [COL] ETU je peux installer/configurer UN connecteur depuis de l'inter-apps](https://trello.com/c/PnUUxk1F/898-5-col-etu-je-peux-installer-configurer-un-connecteur-depuis-de-linter-apps)

manifest.webapp
```diff
@@ -14,11 +14,26 @@
   },
   "default_locale": "en",
   "locales": {},
+  "intents": [
+    {
+      "action": "INSTALL",
+      "data": ["slug"],
+      "href": "/intents",
+      "type": [
+        "io.cozy.konnectors"
```
CPatchane: Should we use a generic `io.cozy.apps` like for the redirection? I mean we could need to install an application from the intent? Or is this completely different? I'd like to keep things generic between konnectors and webapps in the Store.

src/ducks/apps/components/AppInstallation.jsx
```diff
@@ -0,0 +1,102 @@
...
+class AppInstallation extends Component {
+  installApp () {
+    this.setState({ error: null })
+    const { app, onSuccess } = this.props
+    this.props
+      .installApp(app.slug, app.type)
+      .then(() => {
+        onSuccess(app)
+      })
+      .catch()
```
CPatchane: We don't catch anything? Or is this caught by the Redux store? I thinks either we need it and we have to process the error either we don't need it at all here.

src/ducks/components/intents/InstallKonnectorIntent.jsx
```diff
@@ -0,0 +1,113 @@
+import React, { Component } from 'react'
```
CPatchane: Again, has this component to be specific to konnector? Could we use a generic `InstallAppIntent` or this component is doing some things specific to konnectors?

src/ducks/components/intents/IntentHandler.jsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    props.intents
+      .createService()
+    // Free : easy mocking !
+    // Promise.resolve({
+    //   getData: () => ({ slug: <konnector slug> }),
+    //   getIntent: () => ({ action: 'INSTALL', type: 'io.cozy.konnectors' }),
+    //   terminate: (doc) => { alert(`Installed ${doc.name}`) }
+    // })
```
CPatchane: 👍

CPatchane:
Thanks @gregorylegarec 🌜

PR #173, opened 2018-04-19
enhancement: Better IntentHandler, ready to use in any app

gregorylegarec:
An example which speaks by itself:

```jsx
<IntentHandler appData={appData} intents={cozy.client.intents}>
  <InstallAppIntent action="INSTALL" type="io.cozy.apps" />
</IntentHandler>
```


CPatchane:
Thanks @gregorylegarec 🏒

PR #174, opened 2018-04-24
Fix: better handling different icon types

CPatchane:


src/ducks/apps/index.js
```diff
@@ -149,25 +149,32 @@ function _sortAlphabetically (array, property) {
...
   try {
-    icon = await cozy.client.fetchJSON('GET', url)
+    const resp = await fetch(`${COZY_DOMAIN}${url}`, {
+      method: 'GET',
+      credentials: 'include',
+      headers: {
+        Authorization: `Bearer ${COZY_TOKEN}`
+      }
+    })
+    if (!resp.ok) throw new Error (`Error while fetching icon: ${resp.statusText}: ${url}`)
```
gregorylegarec: I think we don't need to throw an error here if we simply ignore it two lines below. Am I right ?

src/ducks/apps/index.js
```diff
@@ -149,25 +149,32 @@ function _sortAlphabetically (array, property) {
   return array.sort((a, b) => a[property] > b[property])
 }
 
+/* Only for the icon fetching */
+const root = document.querySelector('[role=application]')
+const data = root && root.dataset
+const COZY_TOKEN = data && data.cozyToken
+const COZY_DOMAIN = data && (`//${data.cozyDomain}`)
```
gregorylegarec: Is there any other way to avoid dealing with an access to `document` here ? I think that it should be better to instantiate and initialize everything that deal with the `root.dataset` directly in the `index.ejs` file and never hear of it again in the part of the code.

Also, we could pass these data in a prop `appData` and use it as a parameter for `_getIcon(url, appData={})`.

Last option, pass it into react `context` with a top component wrapping everything. For example, the `App` component. See https://medium.com/differential/react-context-and-component-coupling-86e535e2d599.
What do you think ?

src/ducks/apps/index.js
```diff
@@ -149,25 +149,32 @@ function _sortAlphabetically (array, property) {
...
+/* Only for the icon fetching */
+
 async function _getIcon (url) {
   if (!url) return ''
   let icon
   try {
-    icon = await cozy.client.fetchJSON('GET', url)
+    const resp = await fetch(`${COZY_DOMAIN}${url}`, {
+      method: 'GET',
```
gregorylegarec: Don't a `HEAD` call can do the same (I am asking because I am not sure)

src/ducks/apps/index.js
CPatchane: It could be useful for debugging and it was to force using the catch statement here

src/ducks/apps/index.js
CPatchane: Clearly I don't like this usage. But this is particularly due to a cozy-client-js lack, I have to get that data from myself here (or whenever before in the components tree) just because we don't have a `cozy.client.fetch` which just do a `fetch` by adding the token and the domain correctly as we have for `cozy.client.fetchJSON`.

src/ducks/apps/index.js
CPatchane: I don't know, maybe it could work by the `HEAD` is not designed for that according to [the spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4). It's mainly used to get route informations, headers or meta data quickly.

src/ducks/apps/index.js
CPatchane: Could be fixed by the next CozyClient, I added a PR https://github.com/cozy/cozy-client/pull/21
I prefer to keep all here instead of change the application outline just to handle this two "temporary" informations.

src/ducks/apps/index.js
CPatchane: I will remove this throw statement in my next PR ;)

src/ducks/apps/index.js
gregorylegarec: Let's wait for CozyClient.

src/ducks/apps/index.js
gregorylegarec: As we deal directly with the response, `HEAD` is actually not efficient. 👍

PR #175, opened 2018-04-25
Big global improvements

CPatchane:


CPatchane:
[(3) Amélioration Store (Mobile)](https://trello.com/c/y8Bhlcjb/942-3-am%C3%A9lioration-store-mobile)

src/locales/en.json
```diff
@@ -116,6 +116,7 @@
     "partners": "Partners",
     "productivity": "Productivity",
     "ptnb": "expPTNB",
+    "public_service": "Public services",
```
gregorylegarec: Why is the slug singular ?

src/locales/en.json
```diff
@@ -113,9 +113,10 @@
     "insurance": "Insurance",
     "isp": "ISP",
     "others": "Others",
+    "mes_infos": "MesInfos experimentation",
```
gregorylegarec: Shouldn't this kind of category be present in the registry and not in a public app ? We are reproducing the same pattern as Collect with its `konnectors.json`

src/ducks/apps/index.js
```diff
@@ -338,11 +323,17 @@ export function fetchRegistryApps (channel = DEFAULT_CHANNEL) {
...
-                `Something went wrong when trying to fetch more informations about ${
-                  app.slug
-                } from the registry on ${channel} channel, so we skip it. ${err}`
-              )
+              if (err.status === 404) {
+                console.warn(
+                  `No ${channel} version found for ${app.slug} from the registry.`
+                )
+              } else {
+                console.warn(
```
gregorylegarec: Is there a reason here to not throw the Error ? (or at least, to `console.error` it ?)

test/ducks/apps/components/applicationPage/details.spec.js
```diff
@@ -68,15 +68,21 @@ describe('ApplicationPage details component', () => {
   })
 
   it('should handle correctly `display more` behaviour on description part', () => {
-    const component = shallow(<Details {...getAppProps()} />)
+    // window.innerWidth is 1024 by default using jest
```
gregorylegarec: 💡

test/ducks/apps/components/applicationPage/__snapshots__/details.spec.js.snap
```diff
@@ -463,6 +452,16 @@ exports[`ApplicationPage details component should handle correctly \`display mor
...
+Enjoy features to grab your data in you Cozy :tada:### Features
+
+Enjoy features to grab your data in you Cozy :tada:### Features
+
+Enjoy features to grab your data in you Cozy :tada:### Features
+
+Enjoy features to grab your data in you Cozy :tada:### Features
+
+Enjoy features to grab your data in you Cozy :tada:### Features
+
```
gregorylegarec: Is it expected to get five repetitions here ?

src/targets/browser/index.ejs
```diff
@@ -8,6 +8,9 @@
   <link rel="icon" type="image/png" href="/favicon-16x16.png" sizes="16x16">
   <link rel="manifest" href="/manifest.json">
   <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#5bbad5">
+  <% if (__TARGET__ !== 'mobile') { %>
+    <link rel="stylesheet" type="text/css" href="//{{.Domain}}/assets/fonts/fonts.css">
```
gregorylegarec: 👍

src/targets/intents/index.ejs
```diff
@@ -1,22 +1,27 @@
...
-<meta name="viewport" content="width=device-width, initial-scale=1">
-<% _.forEach(htmlWebpackPlugin.files.css, function(file) { %>
-  <link rel="stylesheet" href="<%- file %>">
-<% }); %>
-<% _.forEach(htmlWebpackPlugin.files.js, function(file) { %>
-  <script src="<%- file %>" defer></script>
-<% }); %>
-<% if (__STACK_ASSETS__) { %>
-{{.CozyClientJS}}
-<% } %>
```
gregorylegarec: This was actually a trick by @m4dz to let the browser insert the closing tag by itself.

src/styles/smallAppItem.css
```diff
@@ -13,9 +13,7 @@
   flex-grow: 1;
   flex-shrink: 1;
   flex-basis: calc(100% - var(--item-margin) * 2);
-  width: 20em;
   padding: var(--item-real-margin);
-  /* padding: var(--item-real-margin); */
```
gregorylegarec: Remaining comment

src/locales/en.json
CPatchane: Actually, the store defines a list of authorized categories so the locales should also be here. Unauthorized categoires will be changed to `other` then.

src/ducks/apps/index.js
CPatchane: Yes, this is not an error actually. An application can have an empty list of versions for the default channel (stable here). Just the Store don't find any versions from this channel, it won't be displayed by the store. But it could be useful to be warned about that, couldn't it?

test/ducks/apps/components/applicationPage/__snapshots__/details.spec.js.snap
CPatchane: Yep, for the test I just duplicated the text to reproduce a long text with new lines.

src/targets/intents/index.ejs
CPatchane: Really? What was the trick? Just keep it in the root html tag? I didn't know that behaviour and for me it's clearer with this way.

src/locales/en.json
gregorylegarec: IMHO, it is a role which should be delegated to the registry in the future.

src/targets/intents/index.ejs
gregorylegarec: I think the trick is to save one line of code and a few tabs.

PR #224, opened 2018-06-05
Optimisations + handle realtime when app installing

CPatchane:
Use the brand new [`cozy-realtime`](https://github.com/cozy/cozy-realtime) to handle realtime with the stack.

### DIFF
--- a/src/ducks/apps/index.js
+++ b/src/ducks/apps/index.js
@@ -7,6 +7,7 @@ import constants from 'config/constants'
 import categories from 'config/categories'
 import { extend as extendI18n } from 'cozy-ui/react/I18n'
 import { NotUninstallableAppException } from '../../lib/exceptions'
+import realtime from 'cozy-realtime'
 
 const APP_STATE = {
   READY: 'ready',
@@ -25,11 +26,17 @@ export const REGISTRY_CHANNELS = {
   STABLE: 'stable'
 }
 
+const APPS_DOCTYPE = 'io.cozy.apps'
+const KONNECTORS_DOCTYPE = 'io.cozy.konnectors'
+
 const AUTHORIZED_CATEGORIES = categories
 
 const COLLECT_RELATED_PATH = constants.collect
 
 const DEFAULT_CHANNEL = constants.default.registry.channel
+
+// initial loading
+const LOADING_APP = 'LOADING_APP'
 
 const FETCH_APPS = 'FETCH_APPS'
 const FETCH_APPS_SUCCESS = 'FETCH_APPS_SUCCESS'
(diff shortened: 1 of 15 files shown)


CPatchane:

src/ducks/apps/index.js
```diff
@@ -282,7 +290,73 @@ export function getFormattedInstalledApp(response, collectLink) {
...
+      })
+      .catch(error => {
+        console.warn &&
+          console.warn(`Cannot initialize realtime for apps: ${error.message}`)
+      })
+
+    realtime
+      .subscribeAll(cozy.client, KONNECTORS_DOCTYPE)
+      .then(subscription => {
+        // HACK: the stack creates twice instead of updating
```
gregorylegarec: We should create an issue for this. Did you ?

src/lib/realtime.js
```diff
@@ -0,0 +1,281 @@
+/* global WebSocket */
```
gregorylegarec: I think we may remove this file, summon the wrath of `git commit --fixup` !

src/ducks/apps/index.js
CPatchane: I don't know it it's an issue or a way of working about apps with the stack. Maybe @jinroh can say more about that?

src/lib/realtime.js
CPatchane: Indeed :D

src/ducks/apps/index.js
jinroh: The only difficult thing for us to deactivate is the `create` event sent at the end of the installation, when the application is finally installed and committed into the database.

So simple solution for us is to simply remove the first "fake" `create` event. I think you do not really need it, because it is corresponding to a successful response to the HTTP request starting the installation process.

CPatchane: I think the issue here is the fact of using `create` many times instead of using `update` event, isn't it?

jinroh: Yes, and as I was saying, it is awkward and "difficult" for make the second event an `update` event, since it actually correspond to the creation of the document in database.

Making the first event an `update` event does not really seem like a good idea either ? Or maybe it is ?

CPatchane: Ok I understand. If we can't, I would keep things like that for now. Here we will just duplicate this code to use `onUpdate` as well, not more interesting than just the `onCreate`. If that concerns only the apps/konnectors, I would keep this comment to have the context every time we will read that code :)

CPatchane: Thanks @jinroh for the explanations 👍

PR #225, opened 2018-06-05
Let the registry handle last app versions infos

CPatchane:
Avoid too much requests to fetch app versions, now the registry will return them with the apps listing.
__BREAKING CHANGE :__ This will only work with the latest version of the `cozy-apps-registry`.

### DIFF
--- a/src/ducks/apps/index.js
+++ b/src/ducks/apps/index.js
@@ -316,48 +316,45 @@ export function fetchLatestApp(slug, channel = DEFAULT_CHANNEL) {
   }
 }
 
-export function getFormattedRegistryApp(response, channel) {
-  return cozy.client
-    .fetchJSON('GET', `/registry/${response.slug}/${channel}/latest`)
-    .then(version => {
-      // FIXME retro-compatibility for old formatted manifest
-      const manifest = _sanitizeOldManifest(version.manifest)
-      // source is only used by the stack when installed
-      // so we removed it from the app manifest if present
-      if (manifest.source) delete manifest.source
+export function getFormattedRegistryApp(responseApp) {
+  const version = responseApp.latest_version
+  // FIXME retro-compatibility for old formatted manifest
+  const manifest = _sanitizeOldManifest(version.manifest)
+  // source is only used by the stack when installed
+  // so we removed it from the app manifest if present
+  if (manifest.source) delete manifest.source
 
-      const versionFromRegistry = version.version
-      const screensLinks =
-        manifest.screenshots &&
-        manifest.screenshots.map(name => {
-          const fileName = name.replace(/^.*[\\/]/, '')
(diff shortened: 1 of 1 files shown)


CPatchane:

src/ducks/apps/index.js
```diff
@@ -409,28 +406,21 @@ export function fetchRegistryApps(lang, channel = DEFAULT_CHANNEL) {
...
-                console.warn(
-                  `Something went wrong when trying to fetch more informations about ${
-                    app.slug
-                  } from the registry on ${channel} channel. ${err}`
-                )
-              }
-              return false // useful to skip in an array.map function
+              console.warn(
+                `Something went wrong when trying to fetch more informations about ${
+                  app.slug
```
gregorylegarec: I suggest to use app.name instead for an improved readability.

src/ducks/apps/index.js
```diff
@@ -316,48 +316,45 @@ export function fetchLatestApp(slug, channel = DEFAULT_CHANNEL) {
...
-          }
-        )
-      })
+export function getFormattedRegistryApp(responseApp) {
+  const version = responseApp.latest_version
+  // FIXME retro-compatibility for old formatted manifest
+  const manifest = _sanitizeOldManifest(version.manifest)
+  // source is only used by the stack when installed
+  // so we removed it from the app manifest if present
+  if (manifest.source) delete manifest.source
```
gregorylegarec: Do you think we could put this part inside `_sanitizeOldManifest()` ?

src/ducks/apps/index.js
CPatchane: I would prefer the slug here, it will be better for debugging, won't be? The user can directly fire this message as issue for us.

CPatchane:
We could ensure a legacy, but for now only us are using registries and the store is not released yet. And since the fix is deployed in all registry environments, we don't have to maintain this kind a legacy for almost nothing IMHA. What do you think @gregorylegarec ?

src/ducks/apps/index.js
gregorylegarec: Ok for me.

gregorylegarec:
@CPatchane Ok for me, I was asking for the future. It's pretty sure that this kind of improvements will occur again. In my mind a simple but maybe naive solution could be to expose a registry version.

src/ducks/apps/index.js
CPatchane: Finally nope, we can't because here it's only from registry but this helper is also use when we get apps from the stack. And the source in manifest is used by the stack differently, when the application is installed this source is used to get the channel.
To be clearer, the source should not be provided by the registry in the manifest. If it is, it will replaced by the one used by the stack in all cases.

PR #229, opened 2018-06-06
Load icons after first apps fetching

CPatchane:
Icons fetching is now out from the app initial fetching and a placeholder is displayed while the icone is being fetching after.


CPatchane:

gregorylegarec:
Thanks @CPatchane 👍

PR #235, opened 2018-06-07
Better handling the end of icons fetching

CPatchane:


src/ducks/apps/index.js
```diff
@@ -268,10 +268,12 @@ export function fetchIconsProgressively() {
...
     const iconsMap = new Map() // slug: icon
-    apps.forEach(async (app, index) => {
+    let updateCounter = 0
+    apps.forEach(async app => {
       const iconData = await _getIcon(app.iconToLoad)
       iconsMap.set(app.slug, iconData)
-      if (iconsMap.size % 10 === 0 || index === apps.length - 1) {
+      updateCounter++
+      if (iconsMap.size % 10 === 0 || updateCounter === apps.length) {
         dispatch({ type: RECEIVE_APPS_ICON, iconsMap })
```
ptbrowne: Here, if the update were asynchronous, you would have a problem since the map would be empty when received in the reducer. I think you should dispatch a copy of the map.

src/ducks/apps/index.js
```diff
@@ -268,10 +268,12 @@ export function fetchIconsProgressively() {
...
     const apps = getState().apps.list
     const iconsMap = new Map() // slug: icon
-    apps.forEach(async (app, index) => {
+    let updateCounter = 0
+    apps.forEach(async app => {
       const iconData = await _getIcon(app.iconToLoad)
       iconsMap.set(app.slug, iconData)
-      if (iconsMap.size % 10 === 0 || index === apps.length - 1) {
+      updateCounter++
+      if (iconsMap.size % 10 === 0 || updateCounter === apps.length) {
```
ptbrowne: I don't understand the change :/

CPatchane: As my loop function is asynchronous here, we can't relate on the index argument. So we have to  handle a dedicated counter to know if we are at the end of the update or not.

src/ducks/apps/index.js
CPatchane: Indeed, I may need to improve the map handling a little here 👍

src/ducks/apps/index.js
```diff
@@ -268,12 +268,16 @@ export function fetchIconsProgressively() {
...
       iconsMap.set(app.slug, iconData)
-      if (iconsMap.size % 10 === 0 || index === apps.length - 1) {
-        dispatch({ type: RECEIVE_APPS_ICON, iconsMap })
-        iconsMap.clear()
+      updateCounter++
+      if (iconsMap.size % 10 === 0 || updateCounter === apps.length) {
+        const mapCopy = new Map(iconsMap)
+        await dispatch({ type: RECEIVE_APPS_ICON, iconsMap: mapCopy })
+        // clean already updated icons
+        mapCopy.forEach((v, k) => iconsMap.delete(k))
```
gregorylegarec: We are deleting indexes in `iconsMap` here and at the same time line 274 setting new values in the same map, in a loop with asynchronous method. It looks like a code smell to trigger race conditions and side effects.

The whole code block looks like it could be handled with asynchronous map/reduce functions.

I am ok to merge it as it but we'll need to be very careful with this part.

CPatchane: I see your point. But it seems to work greatly, I didn't noticed issues for now.
I push another commit to compare. I use `for...of` to be sequential but it least it little slower to load every icons.

CPatchane:

gregorylegarec:
Last time we talk about this PR it was still a WIP, tests were ongoing for another icon fetching solution (Use of `Promise.race` or a `<AsyncIcon />` component). Is it still the case ?

src/ducks/apps/index.js
ptbrowne: To me, it's weird to have the promise control flow mixed with the icon logic.

I think it would be better to have something like PromiseChunks that would give you the results in chunks of 10, completely decoupled from the icon logic.

```js
const iconPromises = apps.map(app => _getIcon(app.iconToLoad))
for (let c in chunks(iconPromises, 10)) {
  const chunk = await c()
  await dispatch({ type: RECEIVE_APPS_ICON, chunk })
}
```

More importantly, I wonder why do you chunk in bits of 10 ? Is it to throttle the update of the UI ? Then a temporal throttle on the dispatch function would be clearer I think.

```js
let iconMap = new Map()
const updateIcons = throttle(() => {
  iconMap[res.id] = icon
  dispatch({ type: RECEIVE_APPS_ICONS, iconsMap })
  iconMap = new Map()
}, 100)
const iconPromises = apps.map(app => _getIcon(app.iconToLoad).then(updateIcons))
```

CPatchane: Thanks for these proposals. In fact my real need is to get things progressively and to avoid fetching all these icons in the first app loading. The bunch of 10 fetches is particularly to avoid getting a long time without any icons and let the user see that icons are coming gradually.

CPatchane:
We discussed about this PR with @gregorylegarec, it's clearly perfectible but for now it does the job. Since we need to release the Store very soon and we need this code, we are merging it as is and we will improve this part later. Thanks @gregorylegarec and @ptbrowne for your feedback and help 👍

src/ducks/apps/index.js
gregorylegarec: @ptbrowne Thanks for this other solution. Actually the real issue here is to throttle UI and redux state update. My opinion is that the ideal solution should be to delegate icon fetching and loading to a dedicated component, like <AsyncIcon />, and let React and the browser handle performance concerns.

We'll merge this PR as it as it is just working and we need to release Store and Collect as soon as possible, but we'll keep in mind to update this part of the code.

PR #237, opened 2018-06-07
Add missing doctypes & category

aenario:
Some new (vendor-specific) doctypes were not known, so not clear for the users.

We also added a "onlice_services" category to put the pure-players which are starting to fill up "other"


aenario:
I have the translations ready, let me know when you are ready to merge this, so i can `tx push` :-)

src/locales/en.json
```diff
@@ -117,6 +117,7 @@
     "host_provider": "Host",
     "insurance": "Insurance",
     "isp": "ISP",
+    "online_services": "Services en ligne",
```
CPatchane: I don't see this new one in the `/config/categories.json` it must be added to be completely handled by the Store.

aenario:
fixed, my bad !

PR #242, opened 2018-06-08
feat: add way to bypass default registry channel

ptbrowne:


CPatchane:
Sorry, but we already said that we don't want to add this kind of hack allowing anyone to install apps from unwanted channels on unwanted environments

ptbrowne:
Fine. Keep in mind though that the whole code is open source hence it is somewhat easy to discover that you can target a registry/dev url to install an app in dev mode, and that if this is not controlled server side, nothing will prevent someone willing to do it to install an app from the dev registry.

CPatchane:
@ptbrowne On our environment it won't and it shouldn't be possible to target the Dev channel, where did you see a way to do that? Currently, enabling Dev channel requires an action on server side anyway.

ptbrowne:
> On our environment it won't and it shouldn't be possible to target the Dev channel,

Everything client side can be modified. You can modify the script client side so that the default channel is `dev`.

> Currently, enabling Dev channel requires an action on server side anyway.

Good. Hence "we don't want to add this kind of hack allowing anyone to install apps from unwanted channels on unwanted environments" is not a good reason since it's already taken care of by the server.

Anyway, I think the original point is missed here. My need is to be able to see the description of `dev` apps without the modal in front. I thought that the `localStorage` way was simple and practical since it can be toggled on/off easily even by someone with less technical ability. On Banks, we used this technique to implement [simple flags](https://github.com/cozy/cozy-banks/blob/6ceef1a52ecac2823815b67e38ef35d723cb5738/src/utils/flag.js) that can be turned on for specific cozies. If there is another way you think of that is more practical, I'm all ears :)

CPatchane:
I see your point and your need, we can hack everything on the front side but localStorage is kind of persistant and prone to wrong behaviours or wrong actions from the store on unwanted environments (we could forget a cache).

But we know that the need here is to get a way to see all applications from the Dev channel. We could add something in the future to be able to list the Dev channel according to the context in a dedicated view instead of a hack like this.

PR #246, opened 2018-06-12
Fix alerter messages for install/Uninstall

CPatchane:
Alerter was broken due to cozy-ui update. This PR also add a fix to avoid reacting (page change) on install/uninstall error


CPatchane:
[Je n'ai plus de confirmation lorsque je supprime une application](https://trello.com/c/8tKjje11/1049-je-nai-plus-de-confirmation-lorsque-je-supporime-une-application)

PR #248, opened 2018-06-12
Fix mobileFullscreen prop for IntentModal

gregorylegarec:
Plot twist : case issue.


gregorylegarec:
[Publication Collect/Store](https://trello.com/c/6rT0skFK/1048-publication-collect-store)

PR #250, opened 2018-06-13
Add application vote

CPatchane:
screen shot 2018-06-13 at 14 00 57


CPatchane:
[(1) [STO] ETU je peux voter pour les prochains connecteurs](https://trello.com/c/JUP2cWFL/1038-1-sto-etu-je-peux-voter-pour-les-prochains-connecteurs)

gregorylegarec:
popup.js seems to be imported from Collect, isn't it ? What solution do we have to mutualize it ? Put it in cozy-client ?

CPatchane:
Yes it is. I thought about it and I don't know where to put it :/. Maybe a cozy-libs global, this not related to cozy-client I think.

CPatchane:
@gregorylegarec Now it uses the `PopupOpener` component from cozy-ui 👍

PR #255, opened 2018-06-18
feat: add deezer playlists in doctype locales

doubleface:


src/locales/en.json
```diff
@@ -178,7 +178,8 @@
   "doctypes": {
     "com": {
       "tricount.expenses": "Expenses Tricount",
-      "splitwise.expenses": "Expenses Splitwise"
+      "splitwise.expenses": "Expenses Splitwise",
+      "deezer.playlists": "Deezer playlists"
```
CPatchane: Why playlists specific to deezer instead of a standard playlist? Is it so specific that we have to have an explicit doctype for it?

gregorylegarec:
@CPatchane I think this kind of translation should be in the manifest.konnector, under `deezer.dataTypes.playlist`. What to you think ?

gregorylegarec:
@CPatchane Oh god we are so synchronous.

src/locales/en.json
doubleface: The standard playlist is not defined yet, and we want the connector out. We will standardize a playlist doctype later

doubleface:
@gregorylegarec Can we already do it in the manifest?

src/locales/en.json
CPatchane: What is the format of this doctype? Why can't we take a time to standardise that? At least a first version even if we have to iterate on it later.

gregorylegarec:
@doubleface Yes, we just need to handle it in the store, but all the helpers are already here.

Something like
```js
tOrNothing(`dataTypes.${dataType}`) || tOrNothing(`${konnectorSlug}.dataTypes.${dataTypes}`)
```

aenario:
:warning: it is easier for us if we can translate in the manifrst, but always prioritize translations from this file instead of what is provided by the konnector. (Fishing)

gregorylegarec:
@aenario You're right, but translations from manifest can only override app translation by providing a slug used as a translation prefix : when loading a konnector manifest, all translation are prefixed with the konenctor slug. Example : `trainline.field.login.label`.

src/locales/en.json
aenario: This was decided with most concerned people.

For now every konnector creates its doctype. Once we have 2-3 konnector importing the same kind of data and 1 application using it, we will try to figure out a common `io.cozy.xxxx` doctype and use a service to convert them before migrating the konnectors.

doubleface:
I added the translation in deezer connector manifest

PR #256, opened 2018-06-18
Better install modal UI

CPatchane:
Improve space usage in the modal.

### DIFF
--- a/src/styles/install.css
+++ b/src/styles/install.css
@@ -1,14 +1,13 @@
-.sto-install {
+.sto-install,
+.sto-install-content {
   height: 100%;
   display: flex;
   flex-direction: column;
-  justify-content: space-around;
+  justify-content: space-between;
 }
 
-@media (max-width: 48em) {
-  .sto-install {
-    justify-content: center;
-  }
+.sto-install-content {
+  justify-content: space-between;
 }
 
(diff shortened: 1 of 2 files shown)


CPatchane:
[(1) [Store] Revoir les styles de la modale d'inter apps des permissions](https://trello.com/c/xPZPorYq/1060-1-store-revoir-les-styles-de-la-modale-dinter-apps-des-permissions)

src/styles/install.css
```diff
@@ -17,8 +16,18 @@
   justify-content: center;
 }
 
-.sto-install .sto-install-header {
-  padding-top: 2rem;
+.sto-install-content {
+  overflow-y: scroll;
```
gregorylegarec: Should be `auto` here, otherwise the scrollbar will always be visible.

gregorylegarec:
LGTM except one little change.

gregorylegarec:
Also, there is an empty space on the right

capture d ecran 2018-06-19 a 09 44 15

CPatchane:
@gregorylegarec I don't have this issue in the store, I think this is due to the Intent modal here :/

gregorylegarec:
Thanks @CPatchane 👌

PR #259, opened 2018-06-19
feat: add new doctype com.unibet.bets for bets from unibet

LucsT:


gregorylegarec:
Thanks @LucsT !

@CPatchane I think that this PR should be handled the same way as https://github.com/cozy/cozy-store/pull/255, what do you think ?

LucsT:
@gregorylegarec I let you merge it when you want. And delete the branch. ok?

Thanks

PR #296, opened 2018-07-18
Better handling modals routing with history.replace

CPatchane:
We now use `history.replace` in the ApplicationRouting component instead of `history.push` to handle modals state. It will replace the current history browser state instead of push a new one. We use that to avoid getting back to the uninstall modal (for example) by using the browser "previous button".


CPatchane:
[[STO] Retour en arrière après avoir installer/configuré un connecteur](https://trello.com/c/c0PBGpm6/1124-sto-retour-en-arri%C3%A8re-apr%C3%A8s-avoir-installer-configur%C3%A9-un-connecteur)

PR #306, opened 2018-07-30
Fix broken scroll to top behavior

CPatchane:
Fix scroll to top to automatically scroll to the top of the window when navigating through the application.


CPatchane:
[[STO] Fixer le scrollToTop avec Layout](https://trello.com/c/Myr5iSmz/1129-sto-fixer-le-scrolltotop-avec-layout)

gregorylegarec:
Thanks @CPatchane 💡

PR #319, opened 2018-08-06
Tests improvements

CPatchane:
Update and add some missing tests. Still a lot to do but it's a good step forward.


CPatchane:
[Améliorer les tests du store](https://trello.com/c/ecQNcGZF/1163-am%C3%A9liorer-les-tests-du-store)

PR #339, opened 2018-08-16
Due to cozy-realtime upgrade, the library does not use cozyclient as first parameter

gregorylegarec:
Realtime in store is broken, as cozy-realtime is now intiialized with a config object and not cozy-client anymore.

So in this PR we got the domain and token from the top component, index.jsx.

This is not the best solution imho as the property is passed from component to subcomponents. We should better in the future use a global Cozy object for configuration, or maybe create a `<Cozy />` component which should pass all we need in the context.

### DIFF
--- a/src/ducks/apps/index.js
+++ b/src/ducks/apps/index.js
@@ -413,9 +413,15 @@ function onAppDelete(appResponse) {
 }
 
 function initializeRealtime() {
+  const config = {
+    token: cozy.client._token.token,
+    // cozy-realtime expect an URL with an https protocol,
+    // see https://github.com/cozy/cozy-libs/blob/master/packages/realtime/src/index.js#L52
+    url: `${window.location.protocol}${cozy.client._url}`
+  }
   return async dispatch => {
     realtime
-      .subscribeAll(cozy.client, APPS_DOCTYPE)
+      .subscribeAll(config, APPS_DOCTYPE)
       .then(subscription => {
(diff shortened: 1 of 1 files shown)


ptbrowne:
> This is not the best solution imho as the property is passed from component to subcomponents.

Agreed. I think a singleton object realtime that would configure itself from the HTML directly would be better than passing the props from top to bottom. A context would also be better here, as components in the middle would not know about the config object.

ptbrowne:
This problem seems linked to https://github.com/cozy/cozy-libs/issues/10

gregorylegarec:
@ptbrowne Thanks for the review. At this stage even if we use context, we will pass the values between all components already involved. For example, `index.jsx` will pass domain and token as props to `Root` component, which will pass it through  context but also directly to `<App />`, so we will end with the same cascade. This whole issue will be resolved when https://github.com/cozy/cozy-libs/issues/10 will be implemented.

Also, this is a quick fix to make the master branch work again.

gregorylegarec:
@ptbrowne Ok now, waiting for https://github.com/cozy/cozy-libs/issues/10, I think it's a much better solution.

gregorylegarec:
[[STO] Initialisation du realtime échoue](https://trello.com/c/j5oi8c1J/1171-sto-initialisation-du-realtime-%C3%A9choue)

gregorylegarec:
@ptbrowne If you don't have additional feedback I'll merge today :)

PR #351, opened 2018-08-28
Add new CI workflow

CPatchane:
Use the now CI workflow using cozy-app-publish


CPatchane:
Ok to put in `deploy/` 👍

CPatchane:
We shouldn't to install `cozy-app-publish` since it's already handled by `cozy-scripts` using `cozy-scripts publish`. But cozy-settings don't use `cozy-script` yet.

PR #365, opened 2018-09-04
Fix bug on app channel update

CPatchane:
* Fix channel install not working
* Fix channel detection to display in the app page


CPatchane:
[[STO] La checkbox du changement de channel est cassée](https://trello.com/c/uA8kmP9L/1206-sto-la-checkbox-du-changement-de-channel-est-cass%C3%A9e)

PR #404, opened 2018-10-09
Use the last cozy-scripts with React

CPatchane:
Use cozy-scripts v1 and migrate to React


gregorylegarec:
Thanks @CPatchane ☕️ 

(Remark: most of the commit were more solving styling issue, and could have been something like

> style: Don't use props from render parameter 👕

CPatchane:
Its' not about styling only, using props from render parameter doesn't work on React, it's why this is a fix

PR #409, opened 2018-10-11
Temp fix for back button style

CPatchane:
Add this style to fix a wrong bar styles overwriting


gregorylegarec:
Thanks @CPatchane 👘

PR #410, opened 2018-10-11
Alert on install success

CPatchane:
Was missing from the app installation component

### DIFF
--- a/src/ducks/apps/components/AppInstallation.jsx
+++ b/src/ducks/apps/components/AppInstallation.jsx
@@ -8,9 +8,10 @@ import Spinner from 'cozy-ui/react/Spinner'
 
 import PermissionsList from './PermissionsList'
 import { translate } from 'cozy-ui/react/I18n'
+import Alerter from 'cozy-ui/react/Alerter'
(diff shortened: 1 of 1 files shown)


src/ducks/apps/components/AppInstallation.jsx
```diff
@@ -35,6 +36,11 @@ class AppInstallation extends Component {
       getChannel(prevProps.app.source) === channel
     const succeed = justInstalled || justSwitchedChannel
     if (succeed && typeof onSuccess === 'function') {
```
gregorylegarec: The alert does not depend on the `onSuccess` being a function:
```
if (succeed) {
  if (app.type === APP_TYPE.WEBAPP) {
    Alerter.success(t('app_modal.install.message.install_success'), {
      duration: 3000
    })
 }
 if (typeof onSuccess === 'function') {
   onSuccess
 }
}
```
is better imho.

By the way, `onSuccess` could be validated as function with PropTypes and we could save one test.

gregorylegarec:
Thanks @CPatchane 💂

PR #411, opened 2018-10-12
Hide collect from store apps

CPatchane:
Hide collect since it will be replaced by the Home

### DIFF
--- a/src/config/apps.json
+++ b/src/config/apps.json
@@ -1,9 +1,9 @@
 {
   "notRemovableApps": [
-    "drive",
-    "collect"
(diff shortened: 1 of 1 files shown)


CPatchane:
[(𝐗𝐒) [STORE] Cacher Collect et Home dans l'interface](https://trello.com/c/pLa5jSIm/317-%F0%9D%90%97%F0%9D%90%92-store-cacher-collect-et-home-dans-linterface)

gregorylegarec:
Thanks @CPatchane 👘

PR #412, opened 2018-10-12
Save list scroll position

CPatchane:
A proposal to save the list scroll position when navigating through the store pages:
* If `/<PARENT>/` to `/<PARENT>/<APP>` -> save the scroll position in state
* If `/<PARENT>/<APP>` to `/<PARENT>` -> use the saved scroll position from state
* Other cases, reset scroll position to 0


CPatchane:
[(𝐌) ETU je garde ma position dans la liste des applications](https://trello.com/c/8eFIaUHd/126-%F0%9D%90%8C-etu-je-garde-ma-position-dans-la-liste-des-applications)

PR #425, opened 2018-10-23
fix: Send cookies when fetching manifest.json

nono:
See https://www.w3.org/TR/appmanifest/#obtaining


gregorylegarec:
Thanks @nono 👍

PR #427, opened 2018-10-24
WIP: redux filter

gregorylegarec:
The goal of this PR is to rewrite the filter mechanism of the app to store
the current filter into the redux store.

Previously, the filtering was only based on the parameter in the query string,
which was causing some side effects.

This PR first retrieves the query string and store it into the redux store.

[...]

### DIFF
--- a/src/ducks/components/App.jsx
+++ b/src/ducks/components/App.jsx
@@ -1,3 +1,5 @@
+import 'url-search-params-polyfill'
+
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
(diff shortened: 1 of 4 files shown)


nono:
@gregorylegarec just a friendly reminder that you have this WIP pull request

gregorylegarec:
PR still in pause as this feature priority drastically decreased.

gregorylegarec:
Ok so now this PR is really too old. Let's just keep it as a good a example to handle category filtering directly into the redux store instead of just trusting the URL.

PR #428, opened 2018-10-24
Fix intents

gregorylegarec:
Due to recent changes, intents were broken.


gregorylegarec:
Errored.

ptbrowne:
Maybe add a test ?

PR #437, opened 2018-10-31
Get back scroll behaviour on configure modal

CPatchane:
Update cozy-ui to get scroll locking + only use height for non mobile view.

### DIFF
--- a/src/ducks/apps/components/ConfigureModal.jsx
+++ b/src/ducks/apps/components/ConfigureModal.jsx
@@ -2,6 +2,7 @@ import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import IntentModal from 'cozy-ui/react/IntentModal'
+import withBreakpoints from 'cozy-ui/react/helpers/withBreakpoints'
 
 import { getAppBySlug } from 'ducks/apps'
 
@@ -22,7 +23,8 @@ export class ConfigureModal extends Component {
   }
 
   render() {
-    const { app, dismissAction, onSuccess } = this.props
+    const { app, dismissAction, onSuccess, breakpoints = {} } = this.props
+    const { isMobile } = breakpoints
     return (
       <IntentModal
         action="CREATE"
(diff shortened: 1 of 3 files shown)


CPatchane:
I think this can be considered as a bug from cozy-ui, here is the related issue: https://github.com/cozy/cozy-ui/issues/668

CPatchane:
[[STO/HOME] Je n'ai pas de scroll sur la modale de connexion d'un connecteur](https://trello.com/c/TGHkbqQV/391-sto-home-je-nai-pas-de-scroll-sur-la-modale-de-connexion-dun-connecteur)

src/ducks/apps/components/ConfigureModal.jsx
```diff
@@ -33,7 +35,7 @@ export class ConfigureModal extends Component {
         mobileFullscreen
         overflowHidden
         size="small"
-        height="35rem"
+        height={!isMobile && '35rem'}
```
gregorylegarec: Is there absolutely no way to handle this with CSS ?

CPatchane: We will have to use a static media query matching the mobile breaking point and target a new class on the modal wrapper here.

gregorylegarec: It does not scare me actually and I am pretty much convinced that it is what we should do.

Any opinion about this @GoOz, @ptbrowne ?

ptbrowne: No I think this is OK. It is still a patch.

For me the best way for these problems to go away is for the intent to tell the app what its height ? This way the modal would automatically be the right height.

See my comment : https://github.com/cozy/cozy-ui/issues/668#issuecomment-434641325

GoOz: Well it was discussed in the referenced issue.

gregorylegarec: Ok then, so it’s an interapp issue and we agree on the fact that this magical 35rem value should not be here.

gregorylegarec:
Thanks @Cpatchane 🧞‍♂️

PR #445, opened 2018-11-07
feat: ♿ Add more a11y on Store

GoOz:
Basically:
- Changed some interactive divs into buttons
- Hidden some decorative images from screenreaders
- Added a focus trap when the layer of app details appears in order to make tabulation easier.


GoOz:
[Le Store n’est pas accessible](https://trello.com/c/uZToEApd/119-le-store-nest-pas-accessible)

CPatchane:
Thanks @GoOz ! 🎲

PR #446, opened 2018-11-07
up placeholders in the DOM and create placeholder for application modal

sizvix:
The placeholder was the list apps placeholder , so I made a place holder for the app modal-page
And I move up the place holder in the DOM to not keep them on the first place of render.


sizvix:
[(𝐒) [STO] Attente dans une fiche](https://trello.com/c/82sO2LaP/397-%F0%9D%90%92-sto-attente-dans-une-fiche)

src/ducks/apps/components/Discover.jsx
```diff
@@ -49,6 +49,8 @@ export class Discover extends Component {
     const title = <h2 className="sto-view-title">{t('discover.title')}</h2>
     return (
       <Content className="sto-discover">
+        {isFetching && <OneAppLoading />}
```
CPatchane: I don't understand here, if I understand correctly, we should have the one app loading only when loading the application page, shouldn't we?
It seems like we will have the one app loading on Discover page but apps loading on MyApps page when reading the code

sizvix: That's where you are on an app on the store and press F5

gregorylegarec: The way the fetching is working on cozy-store cannot allow us to distinguish between "are we fetching all the app or only one". However here, if we are fetching all the apps we are fetching the current one as well, so it looks ok to me.

sizvix:
I show a different loading if we are on the list or on an app , using window.location ( like used by routing )

src/ducks/apps/components/ApplicationRouting/AppRoute.jsx
```diff
@@ -2,17 +2,18 @@ import React from 'react'
 import { Route } from 'react-router-dom'
 
 import ApplicationPage from '../ApplicationPage'
+import OneAppLoading from 'ducks/components/OneAppLoading'
 
-export const AppRoute = ({ parent, getApp, isFetching, redirectTo }) => (
+export const AppRoute = ({ parent, getApp, isFetching, redirectTo }) =>{ console.log(parent, getApp, isFetching, redirectTo); return (
```
CPatchane: Oops a console debug line here^^

src/ducks/apps/components/Discover.jsx
```diff
@@ -49,6 +49,8 @@ export class Discover extends Component {
     const title = <h2 className="sto-view-title">{t('discover.title')}</h2>
     return (
       <Content className="sto-discover">
+        {isFetching && !RegExp('(.)*/discover/(.)+').test(window.location) && <AppsLoading />}
```
CPatchane: Why about connecting this component to the router? And then use `location` from `props` it seems cleaner to me. To connecter the router, you can use
```js
import { withRouter } from 'react-router-dom'
...
export default translate()(withRouter(withBreakpoints()(Discover)))
```

and then you will be able to get `location` from `this.props` and do:
```suggestion
        {isFetching && !RegExp('(.)*/discover/(.)+').test(location) && <AppsLoading />}
```

src/ducks/components/OneAppLoading.jsx
```diff
@@ -0,0 +1,134 @@
...
+  style = {
+    ...style,
+    width: widthStyle,
+    height
+  }
+  if (autoMargin) style.margin = 'auto'
+  return <span className="sto-sections-placeholder" style={style} />
+}
+
+export class AppsLoading extends Component {
```
CPatchane: You may want to rename the component to match correctly this file.
Also, could we have a screenshot of this one app loading please? It will be better to see directly what it looks like, thanks :)

sizvix: image

src/ducks/components/OneAppLoading.jsx
```diff
@@ -0,0 +1,134 @@
...
+              <div className="sto-app-info">
+                <div className="sto-app-info-header">Version</div>
+                <div className="sto-app-info-content">
+                  <Placeholder width={[6, 10]} height="1.25rem" />
+                </div>
+              </div>
+              <div className="sto-app-info">
+                <div className="sto-app-info-header">Languages</div>
+                <div className="sto-app-info-content">
+                  <Placeholder width={[6, 10]} height="1.25rem" />
```
gregorylegarec: Personal opinion: I am absolutely not convinced about passing directly style attributes in our components; we should either use classes or use styles, the most important thing should be code consistency.  What should be our guidlines on this point ? ping @ptbrowne @GoOz @Crash-- 

However, it's a task for later out of the scope of this PR, let's keep it as it for now.

gregorylegarec:
(By the way, the first commit should be dropped before merging 👌 )

src/ducks/components/OneAppLoading.jsx
GoOz: I tend to agree. It's bad for consistency and it opens the door for difficult maintenance.

src/ducks/components/OneAppLoading.jsx
```diff
@@ -0,0 +1,134 @@
...
+  style = {
+    ...style,
+    width: widthStyle,
+    height
+  }
+  if (autoMargin) style.margin = 'auto'
+  return <span className="sto-sections-placeholder" style={style} />
+}
+
+export class OneAppLoading extends Component {
```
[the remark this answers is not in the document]
CPatchane: Completely agree about the `<ApplicationPageLoading />` name.
Both solutions are good to me, do the condition before allow us to avoid a render and make the route render directly the loading interface.

src/ducks/apps/components/Discover.jsx
```diff
@@ -49,6 +49,9 @@ export class Discover extends Component {
     const title = <h2 className="sto-view-title">{t('discover.title')}</h2>
     return (
       <Content className="sto-discover">
+        {isFetching &&
+          !RegExp('(.)*/discover/(.)+').test(location) && <AppsLoading />}
```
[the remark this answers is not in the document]
CPatchane: Here we can rely on the props `match.isExact` of our router. If we are on the route `#/discover`, this props will be `true` and if we are on `#/discover/drive` the props will be `false`.

src/ducks/apps/components/ApplicationPage/index.jsx
```diff
@@ -74,12 +75,18 @@ export class ApplicationPage extends Component {
...
+      getApp,
+      redirectTo
     } = this.props
+    if (isFetching) return <ApplicationPageLoading />
+    const app = getApp(matchRoute)
+    if (!app) return redirectTo(`/${parent}`)
+
     const { displayBarIcon } = this.state
     const { isMobile } = breakpoints
     if (isFetching) {
```
CPatchane: Some dead code we can remove since we have this new loading component

CPatchane:
Thanks @sizvix! 🎉

src/ducks/components/Placeholder.jsx
```diff
@@ -0,0 +1,22 @@
...
+    <span
+      className={
+        'sto-sections-placeholder' + (className ? ' ' + className : '')
+      }
+      style={style}
+    />
+  )
+}
+
+export default Placeholder
```
Crash--: Pourquoi il n'est pas dans Cozy-UI ? ;)

gregorylegarec:
Thanks @sizvix ! 👍

src/ducks/components/Placeholder.jsx
gregorylegarec: C'est vrai, nous aurions dû.

sizvix: On a dit qu'il faudrait le faire, mais déjà, valider cette PR qui trainnait, c'était pas mal :P
Et après, il faudrait regarder ce qui est utilisé comme paramètres/options dans les autres placeholders pour les ajouter à cette classe peut-être ¿?

PR #448, opened 2018-11-08
limit the width size of an app modal

sizvix:
To limit the modal element size on huge screen.

### DIFF
--- a/src/styles/application.css
+++ b/src/styles/application.css
@@ -2,6 +2,7 @@
   margin: auto 1.7em;
   height: 100vh;
   background-color: white;
+  max-width: 75rem;
 }
 
 @media (max-width: 48rem) {


src/styles/application.css
```diff
@@ -10,6 +10,12 @@
   }
 }
 
+@media (min-width: 90rem) {
+  .sto-app {
+    width: 1220px; /* White space on the right on huge screen */
```
CPatchane: We usually use `rem` as size unit in our styles. Could you use `75rem` (75*16=1200px) instead?

src/styles/application.css
```diff
@@ -10,6 +10,12 @@
   }
 }
 
+@media (min-width: 90rem) {
+  .sto-app {
+    width: 75rem; /* White space on the right on huge screen */
```
CPatchane: I was wondering if it was not better to use `max-width` instead of `width` here? I would say no with the media query here, but what do you think @GoOz ?

GoOz: Just FYI if you don't want to bothet with rem calculations you can use the mixin `rem(1200)`.

As for the MQ, since it's a `max-width` now, I think it became useless, indeed.

CPatchane: Not in the store since we don't use stylus here^^

GoOz: oh right… 😶

PR #450, opened 2018-11-09
fix: 🐛 Modal works well with FocusTrap

GoOz:
- Add FocusTrap on App installation Modal
- Handle FocusTrap inception

Inception


GoOz:
[[STO] Le Store n’est pas accessible | 1.3.0](https://trello.com/c/uZToEApd/119-sto-le-store-nest-pas-accessible-130)

PR #454, opened 2018-11-12
fix: 🐛 Better app install Modal's layout

GoOz:
- Update UI
- Reforged Modal
- Better Z-Index management


gregorylegarec:
Great job @GoOz, thanks !

PR #456, opened 2018-11-13
Get back animated bar icon on app page mobile

CPatchane:
Now we have to listen the page instead of window since it uses an absolute position


CPatchane:
[(𝐗𝐋) ETU je peux naviguer facilement dans mes applications](https://trello.com/c/iYsSBZio/275-%F0%9D%90%97%F0%9D%90%8B-etu-je-peux-naviguer-facilement-dans-mes-applications)

PR #458, opened 2018-11-13
Intent composition

gregorylegarec:
WIP: should me merge in ~the future `patch-1.2.5`, once~ `patch-1.2.4` ~will be stable~.

CREATE io.cozy.accounts after INSTALL io.cozy.apps

This PR add an interapp composition right after an interapp installation.
This composition is not performed if the first intent is called with a `configure: false` option.


gregorylegarec:
This PR has been moved into [patch-1.2.5](https://github.com/cozy/cozy-store/tree/patch-1.2.5)

PR #463, opened 2018-11-19
fix: 🐛 Tablet layout

GoOz:
- Content page is scrollable again on Tablet
- Content page is well placed on Tablet.

Basically I extended the behaviour for mobile to tablet as well. That was it.


CPatchane:
Thanks @GoOz ! 👍

PR #466, opened 2018-11-21
Improve install intent UX + use non transpiled ui components

CPatchane:
* Redirect to configuration when this is an installed konnector
* Redirect to a success modal if installed application
* Fix some wrong function call since args changing
* Fix spinner position in intents
* __Use back non transpiled cozy-ui components since it's breaking the intents styles and don't work with the transpiled AppIcon__

screenshot 2018-11-22 at 00 56 37


CPatchane:
[🚩 [STO][interapp] L'installation d'une app est répétable à l'infini](https://trello.com/c/7sZ1ClGx/478-%F0%9F%9A%A9-stointerapp-linstallation-dune-app-est-r%C3%A9p%C3%A9table-%C3%A0-linfini)

gregorylegarec:
The first commit should not be in this PR as it fixes the production. At least it should be in a independant PR.

src/ducks/apps/index.js
```diff
@@ -465,45 +467,69 @@ function initializeRealtime() {
   }
 }
 
+async function _getInstalledInfos(app) {
+  try {
+    let installedApp = await cozy.client.fetchJSON(
```
gregorylegarec: This information should already be in the redux store ? I don't see any reason to perform a fetch just to know if an app is installed. But maybe I am missing a point.

src/ducks/apps/components/InstallSuccess.jsx
```diff
@@ -0,0 +1,28 @@
+import React, { PureComponent } from 'react'
+
+import { ModalContent, ModalHeader } from 'cozy-ui/react/Modal'
+import { translate } from 'cozy-ui/react/I18n'
+import AppIcon from 'cozy-ui/react/AppIcon'
+import { fetchIcon } from 'ducks/apps'
+
+export class InstallSuccess extends PureComponent {
```
gregorylegarec: Shouldn't this file be in `src/ducks/components/intents` ? It's only used here.

src/ducks/components/intents/InstallAppIntent.jsx
```diff
@@ -83,6 +84,9 @@ export class InstallAppIntent extends Component {
       null
     )
     const error = !!errorKey && new Error(t(errorKey))
+    const isReadyWithoutErrors = isReady && !error
+    const doConfiguration =
```
gregorylegarec: Repeated code, should be in a function imho. Btw `do` prefix is much a function name prefix. For a variable or a function returning a boolean maybe `hasToBeConfigured` or `isNeedingConfiguration` is better, even a bit longer. What do you think ?

src/ducks/components/intents/InstallAppIntent.jsx
```diff
@@ -34,32 +35,33 @@ export class InstallAppIntent extends Component {
...
+      if (app.type === APP_TYPE.KONNECTOR && doConfiguration) {
         await compose(
           'CREATE',
           'io.cozy.accounts',
           { slug: app.slug }
         )
       }
 
-      onTerminate(nextProps.app)
+      onTerminate(this.props.app)
```
gregorylegarec: `onTerminate` should be call into `AppInstallSucces` now, as it is always displayed. That why we got a termination message too early.

src/ducks/apps/index.js
CPatchane: The intent don't fetch all apps as the main application does. So we have to fetch this information here

src/ducks/apps/components/InstallSuccess.jsx
CPatchane: This component is very generic and could be use to replace the success alert in the main application and there is nothing related to the intents in it

PR #467, opened 2018-11-22
Upgrade cozy-ui to v17

CPatchane:
* Use new fetchIcon props with AppIcon
* Use fixed height for icon (fix Chrome issue when first loading the icon)

### DIFF
--- a/src/ducks/apps/index.js
+++ b/src/ducks/apps/index.js
@@ -195,79 +195,13 @@ function _sortAlphabetically(array, property) {
 /* Only for the icon fetching */
 const root = document.querySelector('[role=application]')
 const data = root && root.dataset
-const COZY_TOKEN = data && data.cozyToken
 const COZY_DOMAIN = data && `//${data.cozyDomain}`
 /* Only for the icon fetching */
 
-const mimeTypes = {
-  gif: 'image/gif',
-  ico: 'image/vnd.microsoft.icon',
-  jpeg: 'image/jpeg',
-  jpg: 'image/jpeg',
-  png: 'image/png',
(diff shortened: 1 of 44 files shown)


CPatchane:
[🚩  [STO] Soucis d'affichage dans Chrome](https://trello.com/c/0S48cZl6/405-%F0%9F%9A%A9-sto-soucis-daffichage-dans-chrome)

src/ducks/apps/index.js
```diff
@@ -195,78 +195,12 @@ function _sortAlphabetically(array, property) {
...
-    icon = new Blob([icon], { type: mimeTypes[extension] })
-  }
-
-  if (icon.type.match(/^image\/.*$/)) {
-    return URL.createObjectURL(icon)
-  }
-  throw new Error(`${app.name}: icon ${url} is not an image.`)
+export const getIconUrl = app => {
+  if (!app || !app.links) return null
+  return `${COZY_DOMAIN}${app.links.icon}`
```
gregorylegarec: This will not work with registry apps

gregorylegarec: As we get the `COZY_DOMAIN` here we are really close to pass it directly to `AppIcon` as prop. `AppIcon` then handles correctly either installed apps and registry apps.

CPatchane: Actually this works with the registry since the Store make the two sources 'iso' with https://github.com/cozy/cozy-store/blob/master/src/ducks/apps/index.js#L562
But I agree, we can make things simpler, I changed it to use directly domain and the secure props 👍

PR #469, opened 2018-11-23
Keep AppIcon height during preload

gregorylegarec:
During preload:

## Before

capture d ecran 2018-11-23 a 18 48 27

## After

capture d ecran 2018-11-23 a 18 48 46


gregorylegarec:
Actually, due to recent update to cozy-ui 17, there is no more need for this.

PR #470, opened 2018-11-23
Fix spinner position on in IntentHandler

gregorylegarec:
Spinner was positionned top right


CPatchane:
Thanks @gregorylegarec! 🍦

PR #481, opened 2018-12-05
Fix channel switching

CPatchane:
Prevent channel switch display if the app is not from the registry or not installed

### DIFF
--- a/src/ducks/apps/components/ApplicationPage/Details.jsx
+++ b/src/ducks/apps/components/ApplicationPage/Details.jsx
@@ -40,20 +40,21 @@ export class Details extends Component {
     this.props.history.push(`/${parent}/${app.slug}/permissions`)
   }
 
-  toggleChannels() {
+  async toggleChannels() {
     const { app } = this.props
-    if (!getChannel(app.source))
-      console.warn("This application don't use the registry")
+    // only for installed apps and apps from the registry
+    if (!(app.installed && getChannel(app.source))) return
+    const stateUpdate = {}
     if (!this.state.displayBetaChannel) {
-      this.setState(() => ({ displayBetaChannel: true }))
+      stateUpdate.displayBetaChannel = true
     }
     if (!this.state.displayDevChannel) {
-      getContext().then(context => {
-        if (context && context.attributes && context.attributes.debug) {
-          this.setState(() => ({ displayDevChannel: true }))
-        }
-      })
+      const context = await getContext()
+      if (context && context.attributes && context.attributes.debug) {
+        stateUpdate.displayDevChannel = true
+      }
     }
+    this.setState(() => stateUpdate)
   }
 
   onUpdateChannel(checked, channel) {
--- a/test/ducks/apps/components/applicationPage/details.spec.js
+++ b/test/ducks/apps/components/applicationPage/details.spec.js
@@ -7,6 +7,7 @@ import { shallow } from 'enzyme'
 
 import { tMock } from '../../../../jestLib/I18n'
 import { Details } from 'ducks/apps/components/ApplicationPage/Details'
+import { REGISTRY_CHANNELS } from 'ducks/apps'
 
 import mockApp from '../../_mockPhotosRegistryVersion'
 import mockKonnector from '../../_mockPKonnectorTrinlaneRegistryVersion'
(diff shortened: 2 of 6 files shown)


CPatchane:
[[STO] Passer sur le channel beta provoque une erreur](https://trello.com/c/jDw5zwhO/551-sto-passer-sur-le-channel-beta-provoque-une-erreur)

src/ducks/apps/components/ApplicationPage/Details.jsx
```diff
@@ -40,20 +40,21 @@ export class Details extends Component {
     this.props.history.push(`/${parent}/${app.slug}/permissions`)
   }
 
-  toggleChannels() {
+  async toggleChannels() {
     const { app } = this.props
-    if (!getChannel(app.source))
-      console.warn("This application don't use the registry")
+    // only for installed apps and apps from the registry
+    if (!getChannel(app.source) || !app.installed) return
```
gregorylegarec: Nitpick:
```
if (!(app.installed && getChannel(app.source))
```
is clearer IMHO and we save a call to `getChannel` it the app is not installed

src/ducks/apps/components/ApplicationPage/Details.jsx
```diff
@@ -40,20 +40,21 @@ export class Details extends Component {
...
-      this.setState(() => ({ displayBetaChannel: true }))
+      stateUpdate.displayBetaChannel = true
     }
     if (!this.state.displayDevChannel) {
-      getContext().then(context => {
-        if (context && context.attributes && context.attributes.debug) {
-          this.setState(() => ({ displayDevChannel: true }))
-        }
-      })
+      const context = await getContext()
```
gregorylegarec: I think we should not get context asynchronously every time we toggle a channel, but load it at application start. Then we could store it:
* in redux store
* in a CozyProvider and pass it to every children component of the application

Moreover, performing an async call on user event can make the whole UX feel slow, especially on mobile and even for a "developer" option.

The whole toggleChannels could be simplified as well:

```jsx
toggleChannels() {
   const { app } = this.props
   const hasAvailableChannels = app.installed && getChannel(app.source)
   if (!hasAvailableChannels) return

   const { showChannels } = this.state
   this.setState({
     showChannels: !showChannels
   })
})
```

and context could be used directly in the render method:

```jsx
  const displayDevChannel = !!this.context.debug
```

Channel beta should be always visible once `this.state.showChannels` is true.

test/ducks/apps/components/applicationPage/details.spec.js
```diff
@@ -67,4 +73,101 @@ describe('ApplicationPage details component', () => {
...
+    }))
+    // we have to re import to handle previous mocking here
+    const MockedDetails = require('ducks/apps/components/ApplicationPage/Details')
+      .Details
+    const appProps = getAppProps()
+    // correctly handled if from registry and installed
+    appProps.app.installed = true
+    appProps.app.source = `registry://${appProps}/stable`
+    const component = shallow(<MockedDetails {...appProps} />)
+    expect(component.state('displayBetaChannel')).toBe(false)
```
gregorylegarec: I think we should not test the implementation (i.e. state attributes).
Asserting the snapshot is correct once the channels have been toggled is enough.

Moreover it's simpler to keep the test consistent when the implementation changes but not the final rendering.

test/ducks/apps/components/applicationPage/details.spec.js
```diff
@@ -67,4 +73,101 @@ describe('ApplicationPage details component', () => {
...
```
gregorylegarec: Same remark.

test/ducks/apps/components/applicationPage/details.spec.js
```diff
@@ -67,4 +73,101 @@ describe('ApplicationPage details component', () => {
...
+          attributes: {
+            debug: true
+          }
+        })
+    }))
+    // we have to re import to handle previous mocking here
+    const MockedDetails = require('ducks/apps/components/ApplicationPage/Details')
+      .Details
+    const appProps = getAppProps()
+    // app not installed
```
gregorylegarec: Every one of these three cases should be an unique test, that's the role of an _unit_ test : test only one condition and expected result.

Same remark about testing the implementation.

gregorylegarec:
Thanks @CPatchane 💡

src/ducks/apps/components/ApplicationPage/Details.jsx
CPatchane: Agree about this change but this is not relevant for this PR since in any cases the context is fetched once then stored in a variable. I would like to have this fetch directly from the client and get this information from the Cozy Provider in the future.

PR #492, opened 2018-12-17
Correctly open install modal if manual update available

CPatchane:
* Fix the issue
* Add tests about this use case


gregorylegarec:
Thanks @CPatchane !

I did not merge immediately after my approval because I wanted to check something.

Actually, I think we should handle the update modal separately, as we already do for the Channel modal.

Adding a conditional route [here](https://github.com/CPatchane/cozy-store/blob/master/src/ducks/apps/components/ApplicationPage/Header.jsx#L79) could do the trick. What do you think ?

CPatchane:
It could do the trick I agree, but since the fix was just a missing check I wouldn't make a refactoring here since the quick fix make the job. But if it doesn't work anymore or if we need more features about here, we will split 👍

PR #496, opened 2018-12-20
Handle app terms and partnership

CPatchane:
This PR is a little big but there all commits are atomic and there were some little cleanings to do.

- Do not check the versions in the registry apps selector (already done on fetching)
- Minor install/uninstall function refactoring to take directly the app object
- Handle terms in the install modal
- Save terms in database as `io.cozy.terms`
- Make class names used by `<ReactMarkdownWrapper` more consistent
- Handle partenship informations in the install modal
- Some minor fixes and cleanings

screenshot 2018-12-20 at 18 39 51


src/ducks/apps/index.js
```diff
@@ -626,29 +630,66 @@ export function uninstallApp(app) {
   }
 }
 
+async function _handleTerms(terms) {
```
ptbrowne: I think the name could be less generic ;)

src/ducks/apps/index.js
```diff
@@ -626,29 +630,66 @@ export function uninstallApp(app) {
...
-      .fetchJSON(verb, `/${route}/${slug}?${queryString}`)
-      .catch(e => {
-        dispatch({ type: INSTALL_APP_FAILURE, error: e })
-        throw e
-      })
+    try {
+      await cozy.client.fetchJSON(verb, `/${route}/${slug}?${queryString}`)
+    } catch (e) {
+      handleError(e)
+    }
```
ptbrowne: No tests :(

src/ducks/apps/components/ApplicationPage/index.jsx
```diff
@@ -9,6 +9,7 @@ import FocusTrap from 'focus-trap-react'
 import AppIcon from 'cozy-ui/react/AppIcon'
 
 import ApplicationPageLoading from 'ducks/components/ApplicationPageLoading'
+import { getTranslatedManifestProperty } from 'lib/helpers'
```
ptbrowne: Nit: I prefer namespaced functions

```
import manifest from 'lib/manifest'

manifest.getTranslatedProperty()
```

CPatchane:
@Crash-- You're right, here it is https://github.com/cozy/cozy-doctypes/pull/70 👍

src/ducks/components/ReactMarkdownWrapper.jsx
```diff
@@ -17,13 +17,18 @@ const parseLinks = props => (
...
-    className="sto-markdown-link"
+    className="md-link"
   >
     {props.children}
   </a>
 )
 
+const parseParagraphs = props => (
+  <p className="md-paragraph">{props.children}</p>
+)
```
ptbrowne: I think it should be renderParagraphs. We are not parsing here.

src/ducks/apps/components/Partnership.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { Component } from 'react'
+import { translate } from 'cozy-ui/react/I18n'
+import { getTranslatedManifestProperty } from 'lib/helpers'
+import ReactMarkdownWrapper from 'ducks/components/ReactMarkdownWrapper'
+
+export class Partnership extends Component {
+  handleBrokenImage = e => {
+    const imgNode = e.nativeEvent.target
```
ptbrowne: I'm surprised that the React event does not have any target ? We cannot check do `const imgNode = e.target` ?

src/ducks/apps/components/Partnership.jsx
```diff
@@ -0,0 +1,38 @@
...
+    if (imgNode) imgNode.style = 'display: none;'
```
ptbrowne: It does not seem very reacty to do this. IMHO it would be less surprising to use a this.setState({ errored: true }) and then checking on this in the render.

src/ducks/apps/index.js
```diff
@@ -251,6 +251,10 @@ function _sanitizeManifest(app) {
     !!sanitized.terms.url &&
     !!sanitized.terms.version
   if (sanitized.terms && !hasValidTerms) delete sanitized.terms
+  // remove incomplete or empty partnership
+  const hasValidPartnership =
+    !!sanitized.partnership && !!sanitized.partnership.description
+  if (sanitized.partnership && !hasValidPartnership) delete sanitized.terms
```
ptbrowne: Maybe we should have helpers like `parternship.isValid` and `terms.isValid` ? Those would be testable ;)

ptbrowne: I see no updated tests from this change :/

src/ducks/apps/index.js
```diff
@@ -282,26 +286,54 @@ export function getContext() {
         })
 }
 
+function _getRegistryAssetsLinks(manifest, appVersion) {
+  if (!appVersion) appVersion = manifest.version
+  const screensLinks =
```
ptbrowne: I think `screenshotLinks` would be clearer. Only 3 character more :)

src/ducks/apps/index.js
```diff
@@ -608,7 +611,10 @@ export function fetchRegistryApps(lang, channel = DEFAULT_CHANNEL) {
   return dispatch => {
     dispatch({ type: FETCH_APPS })
     return cozy.client
-      .fetchJSON('GET', `/registry?limit=150&versionsChannel=${channel}`)
+      .fetchJSON(
+        'GET',
+        `/registry?limit=200&versionsChannel=${channel}&latestChannelVersion=${channel}`
+      )
```
ptbrowne: No tests :/

src/ducks/apps/index.js
```diff
@@ -672,24 +672,33 @@ export function uninstallApp(app) {
 
 async function _handleTerms(terms) {
   const { id, ...termsAttributes } = terms
-  const docId = slugify(`${id}:${termsAttributes.version}`)
+  // We use : as separator for <id>:<version>
+  const docId = `${slugify(id, { remove: /[*+~.()'"!:@]/g })}:${slugify(
+    termsAttributes.version,
+    { remove: /[*+~.()'"!:@]/g }
+  )}`
```
ptbrowne: It's a bit unclear what is the goal here, can we hide this behind a function like `makeTermsKey` or something ?

manifest.webapp
```diff
@@ -56,6 +56,10 @@
     }
   },
   "permissions": {
+    "terms": {
+      "description": "Required to save accepted applications terms",
+      "type": "io.cozy.terms"
+    },
```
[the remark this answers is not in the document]
CPatchane: Since the Store in installed from the platform, there aren't any translations for the permissions descriptions. I can added some here but it won't be used anywhere for now.

src/ducks/apps/index.js
```diff
@@ -192,11 +192,10 @@ function _sortAlphabetically(array, property) {
 /* Only for the icon fetching */
 const root = document.querySelector('[role=application]')
 const data = root && root.dataset
-const COZY_DOMAIN = data && `//${data.cozyDomain}`
 /* Only for the icon fetching */
 
 export const getAppIconProps = () => ({
-  domain: COZY_DOMAIN,
+  domain: data && data.cozyDomain,
```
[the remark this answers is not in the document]
CPatchane: The `<AppIcon />` don't need this `//` at the beginning to works correctly, it was use for the previous way to fetch icons

src/ducks/apps/index.js
```diff
@@ -626,29 +630,66 @@ export function uninstallApp(app) {
...
+        url: termsAttributes.url
+      })
+    }
+  } else {
+    const termsToSave = Object.assign({}, termsAttributes, {
+      _id: docId,
+      accepted: true,
+      acceptedAt: new Date()
+    })
+    await cozy.client.data.create(TERMS_DOCTYPE, termsToSave)
```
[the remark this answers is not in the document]
CPatchane: Nope, the store doesn't for now.

src/ducks/apps/index.js
```diff
@@ -626,29 +630,66 @@ export function uninstallApp(app) {
   }
 }
 
+async function _handleTerms(terms) {
+  const { id, ...termsAttributes } = terms
+  const docId = slugify(`${id}:${termsAttributes.version}`)
```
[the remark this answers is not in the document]
CPatchane: You're right. I was thinking that slugify will also replace special characters but it doesn't, so let use a simple replace with regexp to handle this case and remove it 👍

src/ducks/apps/components/Partnership.jsx
CPatchane: We can't use directly the target property here. React use `SyntheticEvent` which is a layer over the event. Either you persist using `e.persist()` and then use the target or you use the `nativeEvent` property like here:

> [If you find that you need the underlying browser event for some reason, simply use the nativeEvent attribute to get it.](https://reactjs.org/docs/events.html#image-events)

src/ducks/apps/components/ApplicationPage/index.jsx
CPatchane: Not a bad idea but I would keep like this for two reasons (at least for now), we use a lot `manifest` variable name which means the application manifest in the code and we only have one helper relative to the app manifest here.

src/ducks/apps/components/Partnership.jsx
CPatchane: Hum, after thinking I agree, it's better to not rely on the event here 👍

src/ducks/apps/index.js
```diff
@@ -182,10 +182,7 @@ export function getInstalledApps(state) {
 }
 
 export function getRegistryApps(state) {
-  // display only apps with stable versions for now
-  return state.apps.list
-    .filter(app => app.isInRegistry)
-    .filter(app => Array.isArray(app.versions.stable) && !!app.versions.stable)
+  return state.apps.list.filter(app => app.isInRegistry)
```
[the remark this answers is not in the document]
CPatchane: The second filter was unnecessary since the Store already checked for the presence of versions. This filter was just breaking things if we use a channel different from `stable` finally. The tests about this selector still pass (another proof that this filter was useless)

gregorylegarec:
Thanks @CPatchane. Please let's keep testing this PR in the next days.

CPatchane:
@ptbrowne I added a lot of tests for the apps ducks for the reducers and selectors. We just miss some tests for some actions to be 100% tested but this is much better than before 👍

PR #497, opened 2018-12-20
Add a filtered view to get only pending update apps

CPatchane:
Just add the query `pendingUpdate=true` in the URL to get this filtered view

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -29,7 +29,8 @@
         "doctype",
         "slug",
         "tag",
-        "type"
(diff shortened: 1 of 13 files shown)


gregorylegarec:
Too bad I did not have time to finalize this PR https://github.com/cozy/cozy-store/pull/427

CPatchane:
@gregorylegarec I don't think this is necessary (for now at lease) to refactor something that works well and that we don't have issues about

PR #504, opened 2018-12-31
Fix install update

CPatchane:
* Fix application update in install modal
* Correctly remove the available version on update
* Some improvements and tests added

### DIFF
--- a/src/ducks/apps/components/ApplicationRouting/ChannelRoute.jsx
+++ b/src/ducks/apps/components/ApplicationRouting/ChannelRoute.jsx
@@ -21,7 +21,7 @@ export const ChannelRoute = ({
       if (isFetching) return null
       const app = getApp(match)
       const appPath = `/${parent}/${(app && app.slug) || ''}`
-      if (!app || !app.isInRegistry) return redirectTo(`/${parent}`)
+      if (!app) return redirectTo(`/${parent}`)
       const channel = match.params.channel
       const isChannelAvailable = Object.values(REGISTRY_CHANNELS).includes(
         channel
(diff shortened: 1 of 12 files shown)


CPatchane:
[[STO] Je ne peux pas mettre à jour un connecteur manuellement](https://trello.com/c/V2OiF66i/24-sto-je-ne-peux-pas-mettre-%C3%A0-jour-un-connecteur-manuellement)

src/ducks/apps/components/InstallModal.jsx
```diff
@@ -9,10 +9,19 @@ import { hasPendingUpdate } from 'ducks/apps/appStatus'
...
-    const { app, onAlreadyInstalled } = this.props
-
     this.state = { activeTrap: true }
+  }
+
+  componentDidMount = () => {
+    this.handleInstalledStatus()
+  }
+
+  componentDidUpdate = () => {
```
gregorylegarec: When does this modal update ?

src/ducks/apps/components/ApplicationRouting/ChannelRoute.jsx
```diff
@@ -41,7 +41,7 @@ export const ChannelRoute = ({
           isAppFetching={isAppFetching}
           isInstalling={isInstalling}
           onCurrentChannel={() => redirectTo(appPath)}
-          onNotInstalled={() => redirectTo(appPath)}
+          onNotHandled={() => redirectTo(appPath)}
```
gregorylegarec: What does `onNotHandled` mean ?

CPatchane: For the case where the channel modal doesn't not handle this application according to the app state/status

src/ducks/apps/components/InstallModal.jsx
CPatchane: After an install, the modal can be udpdated and we was checking the app status only in the constructor

src/ducks/apps/components/ApplicationRouting/ChannelRoute.jsx
gregorylegarec: It seems that we should not even get there, should we ?

CPatchane: Since we don't check those statuses in the route actually we can get there

PR #505, opened 2018-12-31
Better partner permission text

CPatchane:
Use a generic way to handle partner external permission description

### DIFF
--- a/src/ducks/apps/components/PermissionsList.jsx
+++ b/src/ducks/apps/components/PermissionsList.jsx
@@ -48,7 +48,14 @@ const getProcessedPermissions = (t, app) => {
     const linxoType = 'io.cozy.accounts'
     external.push(
       <Permission
-        description={t('permissions.linxo')}
+        description={
+          app.partnership && app.partnership.name && app.partnership.domain
+            ? t('permissions.banking', {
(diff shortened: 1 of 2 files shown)


CPatchane:
[/!\ Permissions extérieur en dure dans le Store (fait référence à Linxo)](https://trello.com/c/fC38r8nI/58-permissions-ext%C3%A9rieur-en-dure-dans-le-store-fait-r%C3%A9f%C3%A9rence-%C3%A0-linxo)

src/ducks/apps/components/PermissionsList.jsx
```diff
@@ -48,7 +48,14 @@ const getProcessedPermissions = (t, app) => {
...
     external.push(
       <Permission
-        description={t('permissions.linxo')}
+        description={
+          app.partnership && app.partnership.name && app.partnership.domain
+            ? t('permissions.banking', {
+                name: app.partnersip.name,
+                domain: app.partnership.domain
+              })
+            : t('permissions.linxo')
```
gregorylegarec: Is it good to have a locale key named `linxo` ?

CPatchane: Since it's already in the text here I think we can keep it

PR #506, opened 2019-01-02
Handle a previous state on app update and channel switching

CPatchane:
This state handling will avoid to refetch the guessed previous app channel when the user cancel an app update or a channel switching.
Two main new points: 
* Now, if an app update is available, the update modal will fetch the latest app from registry to install it (and overwritte installed app informations) 
* Now, when quitting an install or channel modal, if an application was saved in the state it will be restored into the apps list state. In this case, if the user update an application which was not in the registry it won't be in an incoherent state.

This PR also does some cleaning in props which were not used anymore in the routing.


CPatchane:
[[STO] Je ne peux pas mettre à jour un connecteur manuellement](https://trello.com/c/IjZJvrtG/353-sto-je-ne-peux-pas-mettre-%C3%A0-jour-un-connecteur-manuellement)

src/ducks/components/App.jsx
```diff
@@ -21,6 +21,17 @@ export class App extends Component {
     props.initApp() // fetch apps without icons
   }
 
+  componentDidUpdate(prevProps) {
+    if (!this.props.location) return
+    // quitting channel modal, so we restore the previous app state
+    if (
+      this.props.location.pathname !== prevProps.location.pathname &&
+      prevProps.location.pathname.match(/.*\/channel\/.*/)
```
gregorylegarec: I thought we were clean with logic based on location ? With our recent changes on routing mechanism ? Maybe I am wrong but if not we need a route here.

CPatchane: It's more complicated than just a route here, we need this to detect the action of cancelling an update/channel switching (as discussed)

PR #515, opened 2019-01-17
Handle step on app page in intents

CPatchane:
The need here is to be able to open directly the `install` modal when we use the redirection.
~~I not very happy about the `action` variable naming here but I don't see a very good match here :/~~
We will use the `step` variable to specify `install`, `update`, `permissions` or `uninstall`

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -30,7 +30,8 @@
         "slug",
         "tag",
         "type",
-        "pendingUpdate"
+        "pendingUpdate",
+        "step"
       ]
     },
     {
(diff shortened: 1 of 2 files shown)


CPatchane:
[(𝐒) [Home] En cas de mise à jour disponible, la home redirige vers la modale de mise à jour](https://trello.com/c/DkzSlmgW/802-%F0%9D%90%92-home-en-cas-de-mise-%C3%A0-jour-disponible-la-home-redirige-vers-la-modale-de-mise-%C3%A0-jour)

PR #524, opened 2019-01-22
Improve error message when uninstalling a linked app

nono:
When the user try to uninstall a web application that is linked to a mobile application, an error message is displayed. Before this PR, the error message was a JSON-API body that is not really clear for most users. Now, it is a localized message, with a link to the connected devices page.

#### Checklist

- [ ] I've updated the english locale file, but I've not yet pushed it to transifex (and translating to french needs to be done after that).

### DIFF
--- a/src/ducks/apps/components/UninstallModal.jsx
+++ b/src/ducks/apps/components/UninstallModal.jsx
@@ -17,6 +17,10 @@ export class UninstallModal extends Component {
   constructor(props) {
     super(props)
 
+    this.state = {
+      redirecting: false
+    }
+
     const { app, onNotInstalled } = props
     if (app && !app.installed) {
       onNotInstalled()
@@ -39,6 +43,21 @@ export class UninstallModal extends Component {
     uninstallApp(app)
   }
 
+  toggleRedirect = async () => {
+    if (this.state.redirecting) return // don't toggle twice
+    this.setState(() => ({ redirecting: true }))
+    try {
+      await cozy.client.intents.redirect('io.cozy.settings', {
+        step: 'connectedDevices'
+      })
+    } catch (error) {
+      console.error(error)
+      this.setState({
+        redirecting: false
+      })
+    }
+  }
+
   render() {
     const {
       isUninstalling,
@@ -47,6 +66,9 @@ export class UninstallModal extends Component {
       t,
       uninstallError
     } = this.props
+    const linkedAppError =
+      uninstallError &&
+      uninstallError.message === 'A linked OAuth client exists for this app'
     return (
       <Portal into="body">
         <Modal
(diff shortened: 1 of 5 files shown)


nono:
[(?) ETU mon application web est installée lorsque j'installe mon app mobile](https://trello.com/c/I8RgdcGl/109-etu-mon-application-web-est-install%C3%A9e-lorsque-jinstalle-mon-app-mobile)

src/ducks/apps/components/UninstallModal.jsx
```diff
@@ -40,6 +40,11 @@ export class UninstallModal extends Component {
 
   render() {
     const { isUninstalling, dismissAction, t, uninstallError } = this.props
+    const linkedAppError =
+      uninstallError &&
+      uninstallError.message === 'A linked OAuth client exists for this app'
+    const settingsAppURL =
+      '//' + window.location.host.replace('store', 'settings')
```
nono: I'm not proud of this hack

src/ducks/apps/components/UninstallModal.jsx
```diff
@@ -56,9 +61,19 @@ export class UninstallModal extends Component {
               />
               {uninstallError && (
                 <p className="u-error">
-                  {t('app_modal.uninstall.message.error', {
-                    message: uninstallError.message
-                  })}
+                  {linkedAppError ? (
+                    <>
+                      {t('app_modal.uninstall.linked_app.error') + ' '}
```
nono: I find JSX ugly, with:
- no if/else
- the needs to put `(` and `)` for things on several lines inside `{ ... }`
- `{ ... }` must return a single node, and so I've used the `<> ... </>` hack to regroup the several nodes inside a single one
- I've tried to put an HTML entity to keep the space: `{t('app_modal.uninstall.linked_app.error')} &#32;`, but it was eaten by JSX
- and, in general, so many special characters (``<> {} () ` ` "" '' ?: &&``) in a few lines (when it breaks, it's hard to understand why)

src/ducks/apps/components/UninstallModal.jsx
CPatchane: I think we need a redirection intent to settings here, redirection intent was made exactly for this need

src/ducks/apps/components/UninstallModal.jsx
CPatchane: Was it wanted to keep this link inside the `<p>`? IMHO the good design here was to choose between display the error message in the `<p>` or display a markdown styled component to have the link in the text.
I'll push a change an we'll see 👍

src/ducks/apps/components/UninstallModal.jsx
```diff
@@ -40,6 +40,11 @@ export class UninstallModal extends Component {
 
   render() {
     const { isUninstalling, dismissAction, t, uninstallError } = this.props
+    const linkedAppError =
+      uninstallError &&
+      uninstallError.message === 'A linked OAuth client exists for this app'
```
CPatchane: Is it safe to get an error type based on a description here? Do we have a specific code here or something else with less chance to change

nono: We don't have a specific code for this error. We have `Bad Request` code, but it can be used for other errors.

src/ducks/apps/components/UninstallModal.jsx
nono: > Was it wanted to keep this link inside the `<p>`?

Yes, we want to keep the link inside the `<p>`.

CPatchane:
Thanks @nono 👍

PR #528, opened 2019-01-24
WIP: Bugfixes

CPatchane:
* Use fullwdith for the selectBox
* Remove some css warning about mixed value `start`
* ~~Fix the modal behavior on Safari/iOS which don't allow to use `vh`/`vw` in CSS `calc()`~~
* Fix background issue when opening modals in the app (using Portal)
* Get back permissions icons
screenshot 2019-01-24 at 15 06 18


CPatchane:
[[STO] Reparer les icones des permissions](https://trello.com/c/8eCdWe6q/866-sto-reparer-les-icones-des-permissions)

src/styles/application.css
```diff
@@ -99,7 +99,7 @@ main > [role=main].sto-myapps {
   }
 
   .sto-modal-page-app {
-    height: calc(100vh - 6rem)
+    height: calc(100% - 6rem)
```
gregorylegarec: I am pretty sure it will cause an regression in responsive mode.

CPatchane: This was an issue actually. Safari and Safari iOS don't handle `vh` and `vw` in calc function, this is why we changed that here.

GoOz: Actually not sure we need this change anymore, since we used Portal. Worth checking.

CPatchane: This fis seems not necessary anymore, so I removed the commit

PR #537, opened 2019-02-06
chore: :wrench: Start travis builds only for master and PRs

nono:
Most branches are topic branches that don't need to tested by travis until a PR is made.


CPatchane:
Thanks! 👍

PR #581, opened 2019-03-29
Make apps type filterable and configurable

CPatchane:
Useful to fetch only one or another type of apps


src/config/constants.json
```diff
@@ -5,5 +5,5 @@
     },
     "authorizedLabelLimit": 3
   },
-  "collect": "#/connected"
-}
+  "collect": "#/connected",
+  "filterAppType": false
```
nono: The final `}` has been lost?

src/ducks/apps/index.js
```diff
@@ -411,34 +411,53 @@ export function fetchInstalledApps(lang, fetchingRegistry) {
...
+      let collectLink
+      if (fetchingWebApps) {
+        let installedWebApps = await fetchingWebApps
+        installedWebApps = installedWebApps.map(w => {
+          // FIXME type konnector is missing from stack
+          w.attributes.type = APP_TYPE.WEBAPP
+          return w
+        })
+        // TODO throw error if collect is not installed
+        const collectApp = installedWebApps.find(
```
gregorylegarec: Maybe it's time to rename `collectApp` to `homeApp`

src/ducks/apps/index.js
```diff
@@ -411,34 +411,53 @@ export function fetchInstalledApps(lang, fetchingRegistry) {
...
+          a => a.attributes.slug === 'collect'
+        )
+        collectLink = collectApp && collectApp.links.related
+        installedWebApps = installedWebApps.filter(
+          app => !config.notDisplayedApps.includes(app.attributes.slug)
+        )
+        installedApps = installedApps.concat(installedWebApps)
+      }
+      if (fetchingKonnectors) {
```
gregorylegarec: This block looks pretty much like the previous one, I think we can mutualize it in a dedicated function

src/ducks/apps/index.js
```diff
@@ -456,10 +475,13 @@ export function fetchInstalledApps(lang, fetchingRegistry) {
 export function fetchRegistryApps(lang, channel = DEFAULT_CHANNEL) {
   return dispatch => {
     dispatch({ type: FETCH_APPS })
+    let filterParam = ''
+    if (CONSTANTS.filterAppType)
```
gregorylegarec: Actually `CONSTANTS` is more a global `Config` or `Settings` object. Maybe `storeConfig` ?

src/ducks/apps/index.js
CPatchane: Even better, we don't need this collectApp/Link anymore since we use the intents now, thanks for this point 👍

src/ducks/apps/index.js
CPatchane: For only two lines, it seems overkilled for now, let's keep it like this until the next change here

PR #592, opened 2019-04-04
Add install suggestion service

CPatchane:
Not used for now, but used in customization

### DIFF
--- a/src/targets/services/installSuggestion.js
+++ b/src/targets/services/installSuggestion.js
@@ -1,0 +1,43 @@
+import fetch from 'node-fetch'
+import CozyClient from 'cozy-client'
+
+global.fetch = fetch
+
+const schema = {
+  suggestion: {
+    doctype: 'io.cozy.suggestions'
+  },
+  konnectors: {
+    doctype: 'io.cozy.konnectors'
+  }
+}
+
+let suggestion = {}
+try {
+  suggestion = JSON.parse(process.env.COZY_COUCH_DOC)
+} catch (e) {
+  throw new Error(`Wrong formatted suggestion doc: ${e}`)
+}
+
+const client = new CozyClient({
+  uri: process.env.COZY_URL.trim(),
+  schema,
+  token: process.env.COZY_CREDENTIALS.trim()
+}).getStackClient()
+
+const handleError = (e, slug) => {
+  if (e.status === 409) return
+  throw new Error(`Error when installing ${slug}: ${e}`)
+}
+
+async function install(suggestion) {
+  if (suggestion.silenced) return // silenced, so not concerned
+  const slug = suggestion.slug
+  try {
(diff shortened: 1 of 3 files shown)


CPatchane:
[(𝐒) [STO] ETUO mes connecteurs candidats sont pré-installé](https://trello.com/c/gUFTBk8j/1169-%F0%9D%90%92-sto-etuo-mes-connecteurs-candidats-sont-pr%C3%A9-install%C3%A9)

src/targets/services/installSuggestion.js
```diff
@@ -0,0 +1,42 @@
...
+  token: process.env.COZY_CREDENTIALS.trim()
+}).getStackClient()
+
+async function install(suggestion) {
+  if (suggestion.silenced) return // silenced, so not concerned
+  const slug = suggestion.slug
+  try {
+    await client.fetchJSON(
+      'POST',
+      `/konnectors/${slug}?Source=registry://${slug}/stable`
```
ptbrowne: FMI, is the source not default ?

CPatchane: Good question! I don't know :/
@maximemenager do we have a default source in the stack to install apps/konnectors?

src/targets/services/installSuggestion.js
```diff
@@ -0,0 +1,42 @@
...
+    )
+  } catch (e) {
+    if (e.status === 409) return
+    throw new Error(`Error when installing ${slug}: ${e}`)
+  }
+}
+
+install(suggestion)
```
ptbrowne:
```
const handleErr = err => {
  log('error', err)
}

try {
  install(suggestion).catch(handleErr)
} catch (err) {
  handleErr(err)
}
```

src/targets/services/installSuggestion.js
maximemenager: Yep, if there is no `Source` parameter, the default `Source` the app installer will use the last stable version from the registry

src/targets/services/installSuggestion.js
CPatchane: I don't have any logger for now, so the error handling should be improved but I don't really know how the worker logging system is working for now

src/targets/services/installSuggestion.js
```diff
@@ -0,0 +1,43 @@
...
+  if (suggestion.silenced) return // silenced, so not concerned
+  const slug = suggestion.slug
+  try {
+    await client.fetchJSON('POST', `/konnectors/${slug}`)
+  } catch (e) {
+    handleError(e, slug)
+  }
+}
+
+install(suggestion)
```
ptbrowne: Since the install promise is not awaited, thrown errors (for example if suggestion is null or undefined) will result in UnhandledPromiseRejection, this is why I had put an await and a try catch.

PR #593, opened 2019-04-05
Fix lint

nono:
Use a recent version of yarn on travis to intall the correct version of eslint-preset-cozy-app, and fix the lint now that it doesn't make the build fails.


CPatchane:
Thanks a lot! 🎉

PR #597, opened 2019-04-11
Make store parts configurable + add flag to hide navigation (sidebar)

CPatchane:
* Hide sidebar/filter with `nav=false` as URL query
* Add configuration to display store parts or not

### DIFF
--- a/src/config/index.json
+++ b/src/config/index.json
@@ -6,5 +6,6 @@
     "authorizedLabelLimit": 3
   },
   "collect": "#/connected",
-  "filterAppType": false
+  "filterAppType": false,
+  "enabledStoreParts": ["discover", "myapps"]
 }
--- a/src/config/index.json
+++ b/src/config/index.json
@@ -7,5 +7,5 @@
   },
   "collect": "#/connected",
   "filterAppType": false,
-  "enabledStoreParts": ["discover", "myapps"]
+  "enabledPages": ["discover", "myapps"]
 }
--- a/src/ducks/apps/components/Sections.jsx
+++ b/src/ducks/apps/components/Sections.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import { translate } from 'cozy-ui/react/I18n'
 import withBreakpoints from 'cozy-ui/react/helpers/withBreakpoints'
+import { withRouter } from 'react-router-dom'
 
 import {
   getAppsSortedByCategories,
@@ -10,6 +11,7 @@ import {
 import AppsSection from 'ducks/apps/components/AppsSection'
 import DropdownFilter from 'ducks/apps/components/DropdownFilter'
 import { APP_TYPE } from 'ducks/apps'
+import isNavigationEnabled from 'lib/isNavigationEnabled'
 
 export class Sections extends Component {
   render() {
@@ -21,7 +23,8 @@ export class Sections extends Component {
       allApps,
       query,
       pushQuery,
-      breakpoints = {}
+      breakpoints = {},
+      location
     } = this.props
     const { isMobile, isTablet } = breakpoints
 
@@ -44,10 +47,11 @@ export class Sections extends Component {
     )
 
     const selectOptions = getCategoriesSelections(allApps, t, true)
+    const hasNav = isNavigationEnabled(location.search)
 
     return (
-      <div className="sto-sections">
-        {(isMobile || isTablet) && (
+      <div className={`sto-sections ${hasNav ? '' : '--no-nav'}`}>
+        {(isMobile || isTablet) && hasNav && (
           <DropdownFilter
             options={selectOptions}
             query={query}
@@ -105,4 +109,4 @@ export class Sections extends Component {
   }
 }
 
-export default translate()(withBreakpoints()(Sections))
+export default translate()(withBreakpoints()(withRouter(Sections)))
--- a/src/ducks/apps/components/Sections.jsx
+++ b/src/ducks/apps/components/Sections.jsx
@@ -50,7 +50,7 @@ export class Sections extends Component {
     const hasNav = isNavigationEnabled(location.search)
 
     return (
-      <div className={`sto-sections ${hasNav ? '' : '--no-nav'}`}>
+      <div className={`sto-sections${hasNav ? '' : ' --no-nav'}`}>
         {(isMobile || isTablet) && hasNav && (
           <DropdownFilter
             options={selectOptions}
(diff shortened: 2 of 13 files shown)


CPatchane:
[🔶 (𝐒) ETUO je ne vois que les connecteurs bancaires lorsque j'arrive sur le store filtré](https://trello.com/c/RKVK2rh9/1234-%F0%9F%94%B6-%F0%9D%90%92-etuo-je-ne-vois-que-les-connecteurs-bancaires-lorsque-jarrive-sur-le-store-filtr%C3%A9)

src/ducks/components/Sidebar.jsx
```diff
@@ -1,40 +1,54 @@
...
+    labelKey: 'nav.discover'
+  },
+  myapps: {
+    icon: myAppsIcon,
+    labelKey: 'nav.myapps'
+  }
+}
+
+export const Sidebar = React.memo(({ t }) => {
+  const parts = enabledStoreParts || ['discover', 'myapps']
```
gregorylegarec: I think that enabledStoreParts should be passed as props to the <Sidebar /> component, avoiding repetition of  `enabledStoredPats` in to places of our code.

src/ducks/components/App.jsx
```diff
@@ -16,13 +16,22 @@ import { Layout, Main } from 'cozy-ui/react/Layout'
...
+}
+
 export class App extends Component {
   constructor(props) {
     super(props)
     props.initApp() // fetch apps without icons
   }
 
   render() {
+    const parts = enabledStoreParts || ['discover', 'myapps']
```
gregorylegarec: This information could be passed as props IMHO.

How I see this:
* The `<Root />` component have a dependency to the configuration and inject it into the `<App />` component (it seems logical that an app should have a config property).
* The `<App />` then use `config` to inject props into its children components

Example:
**Root.jsx**
```jsx
<App config={config} />
```

**App.jsx**
```jsx
  render(){
    const { config } = this.props
    const sections = config.enabledSections || ['discover', 'myapps' ]
   // ...
   <Sidebar sections={sections} />
   // ...
   { sections.map(name =>
   // ...
  }
```

src/config/index.json
```diff
@@ -6,5 +6,6 @@
     "authorizedLabelLimit": 3
   },
   "collect": "#/connected",
-  "filterAppType": false
+  "filterAppType": false,
+  "enabledStoreParts": ["discover", "myapps"]
```
gregorylegarec: `sections` looks like a more accurate name to me. `parts` feels too vague. What do you think ?

I don't think we should precise `store` in prop name as we are in store config file.

src/lib/isNavigationEnabled.js
```diff
@@ -0,0 +1,10 @@
+import 'url-search-params-polyfill'
+
+const isNavigationEnabled = (search = '') => {
```
gregorylegarec: Maybe we could merge this file with `lib/getFilteredAppsFromSearch` into a `helpers/navigation`. What do you think ?

src/ducks/apps/components/Sections.jsx
```diff
@@ -105,4 +109,4 @@ export class Sections extends Component {
   }
 }
 
-export default translate()(withBreakpoints()(Sections))
+export default translate()(withBreakpoints()(withRouter(Sections)))
```
Crash--: You should also add some propTypes. Always a good idea ;)

src/ducks/components/Sidebar.jsx
CPatchane: How do you see to pass the config using props here?

src/config/index.json
CPatchane: Sections is already used in the app vocabulary as part of the apps listing :/

src/ducks/components/App.jsx
```diff
@@ -31,10 +40,23 @@ export class App extends Component {
         <Main>
           <Switch>
             <Route path="/redirect" component={IntentRedirect} />
-            <Route path="/discover" component={Discover} />
-            <Route path="/myapps" component={MyApplications} />
-            <Redirect exact from="/" to="/discover" />
-            <Redirect from="*" to="/discover" />
+            {parts.map(name => {
+              if (componentsMap.hasOwnProperty(name)) {
```
Crash--: Is there a reason to use `hasOwnProperty` instead of just `componentsMap[name]`?

src/ducks/components/Sidebar.jsx
```diff
@@ -1,40 +1,58 @@
...
+    labelKey: 'nav.discover'
+  },
+  myapps: {
+    icon: myAppsIcon,
+    labelKey: 'nav.myapps'
+  }
+}
 
-export default translate()(Sidebar)
+export const Sidebar = React.memo(({ location, t, breakpoints = {} }) => {
```
Crash--: 👍

src/ducks/apps/components/Sections.jsx
CPatchane: Yep, I am currently trying to get back this habit to add propTypes :)

src/styles/sections.css
```diff
@@ -2,6 +2,10 @@
   margin: 2rem .5rem 0 2rem;
 }
 
+.sto-sections.--no-nav {
+  margin-top: .5rem;
```
Crash--: No need to create a new class for it. 

`u-mt-half` should be enough, no?

src/config/index.json
CPatchane: I replaced it by `Pages` 👍

src/ducks/components/App.jsx
CPatchane: Using props means that we will have to pass this config props to each children, I am not sure it will be better to undertand :/

src/ducks/components/App.jsx
CPatchane: A reflex, I removed it since we don't need it here 👍

src/lib/isNavigationEnabled.js
CPatchane: `getFilteredAppsFromSearch` is not related to navigation, and it seems clearer to have these helpers in dedicated file since they only do one thing imho

nono:
Ping: can we merge this PR?

src/lib/isNavigationEnabled.js
gregorylegarec: I `getFilteredAppsFromSearch` is actually filtering application based on what is present in the query string, I don't see how we could implement something more related to navigation.

src/ducks/components/App.jsx
gregorylegarec: HOC seems a good idea 👍

gregorylegarec:
Thanks @CPatchane 🔩

PR #598, opened 2019-04-11
wip: Pass provided queries with path redirecting

CPatchane:
Allow to keep the URL queries when navigating within the app

### DIFF
--- a/src/ducks/apps/components/ApplicationRouting/index.jsx
+++ b/src/ducks/apps/components/ApplicationRouting/index.jsx
@@ -22,7 +22,8 @@ export class ApplicationRouting extends Component {
   }
 
   redirectTo = target => {
-    this.props.history.replace(target)
+    const { history, location } = this.props
+    history.replace(target + location.search)
     return null
   }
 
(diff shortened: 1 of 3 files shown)


CPatchane:
[🔶 (𝐒) ETUO je ne vois que les connecteurs bancaires lorsque j'arrive sur le store filtré](https://trello.com/c/RKVK2rh9/1234-%F0%9F%94%B6-%F0%9D%90%92-etuo-je-ne-vois-que-les-connecteurs-bancaires-lorsque-jarrive-sur-le-store-filtr%C3%A9)

CPatchane:
Even with the #427 we will still need to pass them when we navigate through the app, if not they will be removed

gregorylegarec:
I suggested to pass them into the redux store, not into the URL. It needs more dev than just getting what's done in #427.

CPatchane:
So you will have the informations in the URL and also in the redux store, but the redux store should listen the URL to get params update? Because we use the URL to pass those flags, so it seems less simple for the same thing I think

nono:
Ping: should we close this PR?

CPatchane:
This PR is still pending since we need to pass queries to be sure keep them during the application navigation

PR #613, opened 2019-05-15
feat: Use default enabled page 📝

gregorylegarec:
`enabledPages` may be undefined in config, so let set a default one.

### DIFF
--- a/src/ducks/components/App.jsx
+++ b/src/ducks/components/App.jsx
@@ -30,7 +30,7 @@ export class App extends Component {
   }
 
   render() {
-    const defaultPart = enabledPages[0]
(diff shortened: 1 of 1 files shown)


src/ducks/components/App.jsx
```diff
@@ -30,7 +30,7 @@ export class App extends Component {
   }
 
   render() {
-    const defaultPart = enabledPages[0]
+    const defaultPart = enabledPages ? enabledPages[0] : 'discover'
```
CPatchane: The default should be`['discover', 'myapps']` I think, the same than the current Store config.

gregorylegarec: We are just setting `defaultPart` here.

nono:
What should we do about this PR?

PR #619, opened 2019-05-20
feat: Update cozy-client and cozy-realtime

ptbrowne:
No more cozy-client-js and use new cozy-realtime. Done to facilitate
code sharing between applications.

Parts that have changed: 

- reducer des apps
- realtime
- new interapp (bad handshake if we do not use cozy-client v2)

### DIFF
--- a/src/ducks/apps/components/AppInstallation.jsx
+++ b/src/ducks/apps/components/AppInstallation.jsx
@@ -18,6 +18,7 @@ import Checkbox from 'cozy-ui/react/Checkbox'
 import { getTranslatedManifestProperty } from 'lib/helpers'
 import { hasPendingUpdate } from 'ducks/apps/appStatus'
 import storeConfig from 'config'
+import compose from 'lodash/flowRight'
 
 import { APP_TYPE, getAppBySlug, installAppFromRegistry } from 'ducks/apps'
 
@@ -208,7 +209,10 @@ const mapDispatchToProps = dispatch => ({
     dispatch(installAppFromRegistry(app, channel, isUpdate))
 })
 
-export default connect(
-  mapStateToProps,
-  mapDispatchToProps
-)(translate()(AppInstallation))
+export default compose(
+  connect(
+    mapStateToProps,
+    mapDispatchToProps
+  ),
+  translate()
+)(AppInstallation)
--- a/src/ducks/apps/components/AppInstallation.jsx
+++ b/src/ducks/apps/components/AppInstallation.jsx
@@ -21,6 +21,7 @@ import storeConfig from 'config'
 import compose from 'lodash/flowRight'
 
 import { APP_TYPE, getAppBySlug, installAppFromRegistry } from 'ducks/apps'
+import { withClient } from 'cozy-client'
 
 const shouldSkipPermissions = app =>
   flags('skip-low-permissions') &&
@@ -204,12 +205,13 @@ const mapStateToProps = (state, ownProps) => ({
   installError: state.apps.actionError
 })
 
-const mapDispatchToProps = dispatch => ({
+const mapDispatchToProps = (dispatch, ownProps) => ({
   installApp: (app, channel, isUpdate) =>
-    dispatch(installAppFromRegistry(app, channel, isUpdate))
+    dispatch(installAppFromRegistry(ownProps.client, app, channel, isUpdate))
 })
 
 export default compose(
+  withClient,
   connect(
     mapStateToProps,
     mapDispatchToProps
--- a/src/ducks/apps/index.js
+++ b/src/ducks/apps/index.js
@@ -29,23 +29,11 @@ import {
   SAVE_APP
 } from './reducers'
 import { fetchAppsFromChannel, fetchAppOrKonnector } from './client-helpers'
+import { APP_TYPE, APP_STATE, REGISTRY_CHANNELS } from './constants'
 
-const APP_STATE = {
-  READY: 'ready',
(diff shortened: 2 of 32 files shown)


test/ducks/apps/components/applicationPage/__snapshots__/index.spec.js.snap
```diff
@@ -147,7 +147,7 @@ Gestionnaire de photos pour Cozy v3 :tada:",
         namePrefix="Cozy"
         parent="/myapps"
       />
-      <withRouter(Wrapper)
+      <Wrapped
```
CPatchane: Nit for later: we should give a display name to our HOCs for better snapshots reading

src/ducks/apps/index.js
```diff
@@ -240,23 +240,17 @@ function initializeRealtime(client) {
...
-      realtime.subscribe('update', KONNECTORS_DOCTYPE, handleAppUpdate)
-      realtime.subscribe('delete', KONNECTORS_DOCTYPE, handleAppDelete)
-    } catch (error) {
-      // eslint-disable-next-line no-console
-      console.warn(
-        `Cannot initialize realtime for konnectors: ${error.message}`
-      )
+    for (let doctype of [APPS_DOCTYPE, KONNECTORS_DOCTYPE]) {
+      try {
+        realtime.subscribe('created', APPS_DOCTYPE, handleAppUpdate)
```
CPatchane: s/APPS_DOCTYPE/doctype for those three lines

src/ducks/apps/index.js
```diff
@@ -389,42 +390,18 @@ export function fetchInstalledApps(client, lang, fetchingRegistry) {
...
-          return getFormattedInstalledApp(app)
-        })
-      ).then(apps => {
-        return dispatch({ type: FETCH_APPS_SUCCESS, apps, lang })
-      })
+      const shouldBeDisplayed = app => !config.notDisplayedApps.includes(app.attributes.slug)
+      const installedApps = flatten(
+        await Promise.all(promises.filter(Boolean))
+      ).filter(shouldAppBeDisplayed)
+      const apps = await Promise.all(installedApps.map(getFormattedInstalledApp))
```
CPatchane: Do we need a second `Promise.all` here? Since we await just before, we should get the apps here, shouldn't we?

src/ducks/apps/index.js
```diff
@@ -389,42 +390,18 @@ export function fetchInstalledApps(client, lang, fetchingRegistry) {
...
-      Promise.all(
-        installedApps.map(app => {
-          return getFormattedInstalledApp(app)
-        })
-      ).then(apps => {
-        return dispatch({ type: FETCH_APPS_SUCCESS, apps, lang })
-      })
+      const shouldBeDisplayed = app => !config.notDisplayedApps.includes(app.attributes.slug)
+      const installedApps = flatten(
+        await Promise.all(promises.filter(Boolean))
```
CPatchane: For my culture, why do you `filter(Boolean)` here?

src/ducks/components/intents/InstallAppIntent.jsx
```diff
@@ -161,6 +161,7 @@ const mapDispatchToProps = (dispatch, ownProps) => ({
 // translate last to pass the lang property to fetchApps()
 export default compose(
   translate(),
+  withClient,
```
CPatchane: We don't need the client elsewhere in this component? Why do we need the `withClient` here then?

CPatchane: Ok I think this was a duplication, in the changes I see withClient twice

src/ducks/apps/components/AppInstallation.jsx
```diff
@@ -203,12 +205,16 @@ const mapStateToProps = (state, ownProps) => ({
   installError: state.apps.actionError
 })
 
-const mapDispatchToProps = dispatch => ({
+const mapDispatchToProps = (dispatch, ownProps) => ({
   installApp: (app, channel, isUpdate) =>
-    dispatch(installAppFromRegistry(app, channel, isUpdate))
+    dispatch(installAppFromRegistry(ownProps.client, app, channel, isUpdate))
 })
```
Crash--: We had to be careful using `ownProps`. Since it can brings a lot of re-render / re-calculation / dispatch if we have a prop update on the connected component. It can be a big drop of perf. 

Since I'm not familiar with this component, this is just a warning ;)

src/ducks/components/intents/InstallAppIntent.jsx
ptbrowne: Spot on, thx.

src/ducks/apps/index.js
ptbrowne: Thanks !

src/ducks/apps/index.js
ptbrowne: Typo, thanks.

src/ducks/apps/index.js
ptbrowne: `getFormattedInstalledApp` was async so the Promise.all was necessary.

I checked again, and getFormattedInstalledApp does not need to be async so I removed it and removed the Promise.all ;)

test/ducks/apps/components/applicationPage/__snapshots__/index.spec.js.snap
ptbrowne: Good idea, thanks: https://github.com/cozy/cozy-client/pull/440

PR #622, opened 2019-05-21
Refactor Sections

ptbrowne:
In order to extract easily the Sections.jsx component, its responsibility had to be split up.

- The filter functionality was refactored not to rely on the querystring.
- It is now the Sections' responsibility to store the current filter in its state.
- The querystring handling is done by another component that wraps SectionsL QuerystringSections
- The dropdown becomes dumber and knows nothing about the filter
- Discover/MyApplications that used Sections and knew about querystring handling now do not know about it
- Every dependencies of Sections has been put in the same folder for future extraction

Misc: A long standing bug where applications were not properly sorted has been also fixed. I can extract it from this PR if wanted.

Note: The large number of lines changed is a bit skewed because the snapshots contained the list of all apps. One of the prop containing apps has been removed, thus a large number of lines representing apps are no longer there.


src/ducks/apps/components/Sections/components/DropdownFilter.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+import SelectBox from 'cozy-ui/react/SelectBox'
+import Icon from 'cozy-ui/react/Icon'
+import PropTypes from 'prop-types'
+
+const SmallArrow = () => (
+  <Icon
+    className="sto-sections-dropdown-icon"
+    icon="small-arrow"
+    color="#95999D"
```
Crash--: I know this is not part of this PR but : coolGrey ;)

src/ducks/apps/components/Sections/Sections.jsx
```diff
@@ -0,0 +1,167 @@
...
+      </div>
+    )
+  }
+}
+
+Sections.propTypes = {
+  t: PropTypes.func.isRequired,
+  apps: PropTypes.array.isRequired,
+  error: PropTypes.object,
+  onAppClick: PropTypes.func.isRequired
```
Crash--: hasNav: PropTypes.bool.isRequired

ptbrowne: Well spotted, I did not put the required since hasNav is in the defaultProps ;)

src/ducks/apps/components/Sections/components/DropdownFilter.jsx
ptbrowne: Done in https://github.com/cozy/cozy-store/pull/622/commits/b1367b25992e4d1363f01db9adbc6f91bd33c625 (`var(--coolGrey)`).

src/ducks/apps/components/Sections/components/AppsSection.jsx
```diff
@@ -18,6 +18,8 @@ const _renderAppComponent = (app, t, onAppClick, isMobile) => {
   )
 }
 
+const appSorter = t => app => getTranslatedManifestProperty(app, 'name', t)
```
CPatchane: nit: I would rename this `makeNameGetter` since this is clearly returning a getter and not a sorter here

src/ducks/apps/components/Sections/components/AppsSection.spec.jsx
```diff
@@ -44,6 +44,6 @@ describe('AppsSection component', () => {
       .dive() // shallow on more level on first app item
     appItem.simulate('click')
     expect(mockOnAppClick.mock.calls.length).toBe(1)
-    expect(mockOnAppClick.mock.calls[0][0]).toBe(mockApps[0].slug)
+    expect(mockOnAppClick.mock.calls[0][0]).toBe('konnector-bouilligue')
```
CPatchane: Why not use the mock object directly? It avoids hard written strings twice in tests

src/ducks/apps/components/Sections/components/DropdownFilter.jsx
```diff
@@ -8,7 +8,7 @@ const SmallArrow = () => (
   <Icon
     className="sto-sections-dropdown-icon"
     icon="small-arrow"
-    color="#95999D"
+    color="var(--coolGrey)"
```
CPatchane: I didn't know this way was working with this component 👍

src/ducks/apps/components/Sections/categories.js
```diff
@@ -57,14 +57,18 @@ export const addLabel = (cat, t) => ({
...
  * Returns unlabelised/unsorted categories from a list of apps
  * @param  {Array<App>}  apps
  * @param  {Boolean} includeAll - Whether to add an "All" option
  * @return {Array<CategoryOption>}
  */
-export const generateOptionsFromApps = (apps, includeAll = false) => {
+export const generateOptionsFromApps = (
+  apps,
+  includeAll = false,
+  _addLabel
```
CPatchane: Maybe `includeAll` and `_addLabel` should be in an `options` object? They are optional, and it will be better for code reading when using this function: `generateOptionsFromApps(apps, { includeAll: true })`

src/ducks/apps/components/Sections/components/AppsSection.spec.jsx
ptbrowne: Because the apps are now sorted, it is not necessarily the first one.

Do you think it would be clearer with 

```
     expect(mockOnAppClick.mock.calls[0][0]).toBe(mockApps[6].slug)  // IDK the real index of konnector-bouilligue but it is not 0
```

?

CPatchane: Ok I see, I understand now, let's keep it like thi.

src/ducks/apps/components/Sections/components/AppsSection.jsx
ptbrowne: Done in https://github.com/cozy/cozy-store/pull/622/commits/5636fdb6108fedb74b5051ec0970fded22f4ef27

src/ducks/apps/components/Sections/categories.js
ptbrowne: Done in https://github.com/cozy/cozy-store/pull/622/commits/5636fdb6108fedb74b5051ec0970fded22f4ef27

PR #656, opened 2019-09-02
chore: Fix cozyPublish script

drazik:
There is a problem with cozy-scripts' publish script that prevents us
from deploying the store to the registry. See
https://travis-ci.org/cozy/cozy-store/jobs/578391579#L570. This looks a
hard to debug issue. Using npx while we investigate this issue ensures
that the deployment is still working.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "watch:browser": "cozy-scripts watch --browser",
     "watch:mobile": "cozy-scripts watch --mobile",
     "start": "cozy-scripts start --hot",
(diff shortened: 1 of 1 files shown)


drazik:
> It doesn't work with `yarn cozy-app-publish` ?
> 
> So the issue is not in cozy-scripts but really in cozy-app-publish ?

`cozy-app-publish` is not in the dependencies because `cozy-scripts` installs the latest version globally before running it (which should not be the case IMHO, but it's another issue). It seems this global install fails on travis for a reason I don't find. The best way to do the same thing without our custom code to install the package globally is to use `npx`.

PR #660, opened 2019-09-03
docs: Fix link to *Install the development env.*

sebn:
Actually link to the *Prerequisites* section so the reader can see both of them.


Crash--:
Thanks @sebn !

sebn:
Thanks for the fast merge @Crash-- !

Also please note the `yarn run watch:server` documented in README doesn't exist anymore.
It could probably be replaced with `yarn start`, in which case the whole *Run it inside the VM* section may not be relevant anymore, but I'm not sure :wink:

PR #673, opened 2019-11-07
feat: Add io.cozy.apps.suggestions

ptbrowne:
- Apps suggestions permission is shown with the apps icon.
- Add locales

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "cozy-bar": "7.7.5",
     "cozy-client": "6.64.0",
     "cozy-flags": "1.10.0",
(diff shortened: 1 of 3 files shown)


src/locales/en.json
```diff
@@ -214,6 +214,7 @@
       "cozy": {
         "accounts": "Login details",
         "apps": "List of installed applications",
+        "apps.suggestions": "Suggestions of applications",
```
y-lohse: Nit: I would say "Application suggestions" sounds more natural, but I don't have the whole context.

ptbrowne: Thanks, fixed.

PR #684, opened 2020-05-12
feat: Add recurrence doctype support

ptbrowne:
- [x] src/locales/en.json
- [x] src/config/permissionsIcons.json
- [x] src/assets/icons/permissions/bank-recurrence.svg

@joel-costa I have made the icon for the bank recurrence doctype, tell
me if you want to change it.


Crash--:
Did you update the file on transifex too?

ptbrowne:
Ah I had changed the fr.json, and then saw that it was not picked up by the commit, and then forgot to go to Transifex to update it...

As we discussed before about putting Transifex files in the repo, I did that:  

- removed locales files from .gitignore and added them to the repo
- added the recurrence string in the French translation
- pushed on transifex

ptbrowne:
Good point. The nl seems OK though, I kept it.

I had forgotten to remove transifex from the Travis config file, it' done now.

PR #692, opened 2020-09-17
feat: Add missing doctypes for permission page

ptbrowne:
- Missing locale
- Missing icon

for io.coy.bank.accounts.stats and io.cozy.bank.recipients


Crash-- (approved):
trop bien ! Enfin :D

PR #694, opened 2020-09-18
Add search input

ptbrowne:
This adds as a search input that filters the apps and konnectors
client-side via fuse.js

It also refactors the Sections since the AppsSection in cozy-ui
already does the job. This way, the store in Banks is the same
as the store (Banks already used the component that had been
extracted from the store).


src/ducks/apps/components/Sections/Sections.jsx
```diff
@@ -85,9 +85,12 @@ export class Sections extends Component {
     super(props, context)
     this.state = {
       filter: {},
-      searchFieldValue: '',
-      filteredApps: this.getFilteredApps()
+      searchFieldValue: ''
     }
+
+    // getFilteredApps must have state.filter already ready
+    this.state.filteredApps = this.getFilteredApps()
```
Crash--, line 92:
```suggestion
this.state = { 
filter: {},
searchFieldValue: '',
// getFilteredApps must have state.filter already ready
filteredApps: this.getFilteredApps()
}
```

I found it clearer

src/ducks/apps/components/Sections/Sections.jsx
```diff
@@ -108,110 +109,96 @@ export class Sections extends Component {
...
-    // search is taken from props or state
-    const search = this.props.search || this.state.search
-    return search
+  componentDidUpdate(prevProps) {
+    if (
+      prevProps.filter !== this.props.filter ||
+      prevProps.apps !== this.props.apps
+    ) {
+      this.updateFilteredApps()
+      this.setState({ searchResults: null, searchFieldValue: '' })
```
Crash--, line 132: doesn't that mean you remove the value of the searchField input when a search is done? 

I have difficulty understanding this part

src/ducks/apps/components/Sections/Sections.jsx
ptbrowne: This cannot work as `this.state` is declared after `this.getFilteredApps()` runs but `this.getFilteredApps` relies on `this.state`. This is why there is a comment. I could replace "already ready" by "already declared", I think this would make it clearer.

src/ducks/apps/components/Sections/Sections.jsx
ptbrowne: The apps are coming from the above, the filter is done on the categories of apps, and is modified for example when switching categories either by going to a different location via the nav or by selecting another category option. I think I'll add a docstring explaining the differnce between the filter (from the props), and the search (from the state) in the docstring of Section to make it clearer. What do you think ? Does it make more sense to you ?

src/ducks/apps/components/Sections/Sections.jsx
Crash--: Also, you can't set this `filteredApps` on did/WillMount() for instance?

ptbrowne: > getFilteredApps can also does something

Yes it can. I envisioned this possibility and prefer the other way.

> Also, you can't set this filteredApps on did/WillMount() for instance?

I don't understand what you mean to say by that ? Do I need to do something ?

src/ducks/apps/components/Sections/Sections.jsx
ptbrowne: Done in https://github.com/cozy/cozy-store/pull/694/commits/e2e493b4355d89dff90020daca01c63836e064fb

src/ducks/apps/components/Sections/Sections.jsx
Crash--: Instead of calling getFilteredApps in the constructor and dealing with the fact that state.filter should have been instanciated, you can call getFilteredApps in the willmount. Like that, you're sure that state.filter is already declared

ptbrowne: I am not sure what problem we are solving here ? `componentWillMount` is deprecated

I do no think that
```
componentWillMount() {
  this.setState({ filteredApps: this.getFilteredApps() })
}
```

is better than the original code.

Crash--: didMount does the trick without using deprecated method. 

By doing so, we solve the issue when reading this code, not understanding why `state.filteredApps` is managed differently than the others and having to read the comment and the `getFilteredApps()` code to understand why. 

But never mind

src/ducks/apps/components/Sections/index.spec.jsx
```diff
@@ -3,111 +3,85 @@
...
-      />
-    ).getElement()
-    expect(component).toMatchSnapshot()
+  it('should filter the results', async () => {
+    const { root } = setup()
+    const input = root.getByPlaceholderText('"SNCF", "telecom", "bills"')
+    act(() => {
+      fireEvent.change(input, { target: { value: 'Bouil' } })
+    })
+    await sleep(500)
```
Crash--, line 77: Why 500? Because of the debounce (now at 300) ?

Crash-- (approved):

src/ducks/apps/components/Sections/Sections.jsx
ptbrowne: Yep, but doing it in didMount will incure another render, this is why I had done in the constructor and I will change the comment so that it's clearer.

src/ducks/apps/components/Sections/index.spec.jsx
ptbrowne: Yes. I changed the code to use a mocked debounce to circumvent this problem (so that the tests are faster). Jest fake timers do not work with debounce unfortunately.

Crash--: I love the idea of mocking the debounce! <3

PR #695, opened 2020-09-18
Improvements

ptbrowne:
Small improvements and tweaks made during the search 
implementation. I've splitted the PR in two to facilitate
merging of things that is not directly part of the
search implementation. Commit messages should be self
explanatory.


src/lib/store.js
```diff
@@ -13,7 +14,12 @@ const loggerMiddleware = createLogger()
 
 const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose
 
-const middlewares = [thunkMiddleware, loggerMiddleware]
+const middlewares = [
+  thunkMiddleware,
+  flag('store.redux-logger') && process.env.NODE_ENV !== 'browser:production'
+    ? loggerMiddleware
+    : null
+].filter(Boolean)
```
Crash--, line 22: this is not a `chore` commit since you are changing the code we use in production ;)

Why this test on NODE_ENV?

Crash-- (approved):
tests are red since you've a wrong import. but 👍

PR #705, opened 2020-12-04
fix: Some hidden connectors in store

doubleface:
This is due to a hardcoded limit of 200 in the request getting the list
of apps and connectors.
I set it to 300 for the weekend, but the better fix would be to add
registry to cozy-client and with a proper pagination handling.


Crash--:
FYI @doubleface we already have the `registryCollection` in cozy-client : https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/registry.js#L78 

You should be able to use it for now. Then we can enhance it to handle pagination

PR #711, opened 2021-03-01
feat: Add permission icon for geojson doctype

paultranvan:
Add a location icon for geojson doctype, which will be useful for the tracemob konnector


.travis.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         - if [ "$TRAVIS_SECURE_ENV_VARS" != "false" ]; then eval "$(ssh-agent -s)"; fi
         - if [ "$TRAVIS_SECURE_ENV_VARS" != "false" ]; then chmod 600 /tmp/id_rsa_downcloud_cozy-store; fi
         - if [ "$TRAVIS_SECURE_ENV_VARS" != "false" ]; then ssh-add /tmp/id_rsa_downcloud_cozy-store; fi
-        - curl -fsSL https://bootstrap.pypa.io/get-pip.py | python - --user
+        - curl -fsSL https://bootstrap.pypa.io/2.7/get-pip.py | python - --user
```
ptbrowne: Why do we need pip, and why do we need at 2.7 since there is no support for Python 2.7 anymore ?

ptbrowne: I think we do not need pip, it must be a remnant of the need for transifex. Can you try to remove it ?

Crash--: done in the next commit ;)

Crash-- (approved):

PR #744, opened 2021-08-25
feat: Reword permissions modal

Ldoppea:
Permission modal was not clear enough about what was shared or not

We think that the new wording will fix this problem

### Before:
image

### After:
image


Ldoppea:
Note that the distinction between `internal` (when there are also external permission) and `internal_only` (when there is no other external permission) is disappearing. I'm waiting for some answers from product-owners but I may merge those two translation entries later.

acezard (approved):

PR #747, opened 2021-08-26
fix: Make the main content fill the remain space of the browser view

Ldoppea:
⚠️ this PR is a draft as it should be completed by bumping `cozy-ui` version after related PR is merged

When the main content (application list) is scrollable but its content is not large enough (i.e. a few apps in each list section), then the scroll bar is stick to the apps list instead of being stick to the browser right side

This fix make the main content always take all remaining space

`u-flex-grow-1` has been added to `<CozyTheme>` as this is the first child directly under the main application container (which is set to be a flexbox)

Also the scrollable area highly depends on navbar size, this has been fixed in cozy-ui with https://github.com/cozy/cozy-ui/pull/1881 

### Before:
image

### After:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "cozy-interapp": "0.6.2",
     "cozy-logger": "1.3.1",
     "cozy-realtime": "3.10.5",
-    "cozy-ui": "^50.1.2",
+    "cozy-ui": "^51.7.1",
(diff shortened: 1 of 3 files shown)


Ldoppea:
This PR is now ready for review

Merkur39 (approved):

PR #770, opened 2022-01-25
feat: use cozy-intent instead of konnectorOpenUri

Ldoppea:
Old `konnectorOpenUri` mechanisme is removed in favor of cozy-intent
based one

Those mechanisms are used to tell parent react-native app that a
connector pane needs to be opened

`konnectorOpenUri` required the parent react-native app to inject a
token in the store's url so the store would know if executed in a
react-native context or not, and then intercept navigation URLs to
detect if a connector needs to be opened. This implementation was not
easy to understand

New implementation uses cozy-device-helper to detect if run from
CozyFlagship app and the uses cozy-intent messaging service to ask for
displaying a connector

---

TODO:

- [x] Update Cozy-intent to retrieve dependency injection from https://github.com/cozy/cozy-libs/pull/1438

### DIFF
--- a/src/ducks/apps/components/ApplicationPage/Header.jsx
+++ b/src/ducks/apps/components/ApplicationPage/Header.jsx
@@ -9,6 +9,8 @@ import { translate } from 'cozy-ui/transpiled/react/I18n'
 import withBreakpoints from 'cozy-ui/transpiled/react/helpers/withBreakpoints'
 import Intents from 'cozy-interapp'
 import { withClient } from 'cozy-client'
+import { useWebviewIntent } from 'cozy-intent'
+import { isFlagshipApp } from 'cozy-device-helper'
 
 import cozySmileIcon from 'assets/icons/icon-cozy-smile.svg'
 import AsyncButton from 'ducks/components/AsyncButton'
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -46,9 +46,10 @@
     "cozy-app-publish": "^0.27.2",
     "cozy-bar": "7.16.0",
     "cozy-client": "22.1.0",
-    "cozy-device-helper": "1.12.0",
+    "cozy-device-helper": "1.16.0",
     "cozy-doctypes": "1.72.2",
     "cozy-flags": "2.6.0",
+    "cozy-intent": "^1.3.0",
```
trollepierre, line 52: should we use fixed dependencies or should we use ^ ?

the PR does not look consistent in my opinion.

src/ducks/components/App.jsx
```diff
@@ -37,34 +38,36 @@ export class App extends Component {
...
-                <Fragment>
-                  <Redirect exact from="/" to={`/${defaultPart}`} />
-                  <Redirect from="*" to={`/${defaultPart}`} />
-                </Fragment>
-              )}
-            </Switch>
-          </Main>
-          <IconSprite />
-        </Layout>
+        <WebviewIntentProvider>
```
trollepierre, line 41: i don't think the PR is ready for review. Tests are failing :(

trollepierre (changes requested):

src/ducks/components/App.jsx
Ldoppea: Yes, I did not edit tests yet. Those are failing because one test is still checking for old mechanism.

The new implementation can still be reviewed for feedback, it should not change a lot.

package.json
Ldoppea: Fixed

src/ducks/apps/components/ApplicationPage/Header.jsx
```diff
@@ -29,25 +31,35 @@ export const Header = ({
...
+  }
+  const openConnector = () => {
+    if (isFlagshipApp()) {
+      return webviewIntent.call('openApp', app.related, app)
+    } else {
+      const intents = new Intents({ client })
+      return intents.redirect('io.cozy.accounts', {
+        konnector: app.slug
+      })
+    }
```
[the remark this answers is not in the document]
Ldoppea: Fixed in https://github.com/cozy/cozy-store/pull/770/files#diff-a28fa8a6e0b5386b123c61675c3460874a2411bd4a98df08dbb5a4f738ec6691

src/ducks/components/App.jsx
Ldoppea: Tests are implemented and CI is green 👍

trollepierre (approved):

PR #782, opened 2022-04-04
feat: Add coachco2 settings locales

JF-Cozy:


src/locales/en.json
```diff
@@ -242,6 +242,7 @@
         "bills": "Bills",
         "certified.carbon_copy": "Carbon copy",
         "certified.electronic_safe": "Electronic safe",
+        "coachco2.settings": "CoachCO2 app settings",
```
Ldoppea, line 245: We have `Bank app settings` but also `Photo settings` maybe someday we will have to homogenise those 😛

Ldoppea (approved):

PR #784, opened 2022-04-27
feat: Add cc.cozycloud.errors permission

Crash--:
Since `cc.cozycloud.sentry` is deprecated, we are migrating to `cc.cozycloud.errors`. So let's add this permission


trollepierre (approved):
I didn't get the end of the PR description. 🤣

PR #785, opened 2022-04-28
refactor: Fix Proptype issue

Crash--:
We were passing `false`

### DIFF
--- a/src/ducks/apps/components/ConfigureModal.jsx
+++ b/src/ducks/apps/components/ConfigureModal.jsx
@@ -42,7 +42,7 @@ export class ConfigureModal extends Component {
         mobileFullscreen
         overflowHidden
         size="small"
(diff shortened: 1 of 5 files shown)


src/ducks/apps/components/ConfigureModal.jsx
```diff
@@ -42,7 +42,7 @@ export class ConfigureModal extends Component {
         mobileFullscreen
         overflowHidden
         size="small"
-        height={!isMobile && '41rem'}
+        {...(!isMobile ? { height: '41rem' } : {})}
```
trollepierre, line 45: nit: is it possible to test this height is not set to false anymore?

trollepierre (approved):
let's fix the pipeline 🙏

Ldoppea:
@Crash-- seems like the auto-merge bot did not do its job here.

PR #799, opened 2022-07-18
fix: Call Flagship app's intent after installing a connector

Ldoppea:
After installing a connector, `cozy-store` calls `cozy-home` intent
that is displayed inside of an iFrame

When rendered inside of Flagship app we don't want to do this
redirection but instead we want to call the native intent to open the
connector from the Home view


trollepierre (approved):

Ldoppea:
I updated the PR description and title and force pushed commits.

I now have 2 commits that are mutually exclusive, I would like your thoughts about them.

[1st commit](https://github.com/cozy/cozy-store/pull/799/commits/3981105e035e66df8f1bdd667d4407ef81b6d2ed) implements the fix from `InstallRoute.jsx` and changes the `redirectToConfigure` behaviour: 

[2nd commit](https://github.com/cozy/cozy-store/pull/799/commits/18bce6e7d2217dd6fcfa74875b73b1f6426aaa0e) implements the fix from `InstallRoute.jsx` and `AppInstallation.jsx`. It is closer from what had been done in [this old commit](https://github.com/cozy/cozy-store/commit/91a3335b09cded0c9b831d470476033ec6b62589). But I think it changes more the concepts behind `onInstallOrUpdate()`. This seems ok but I'm not sure why `onInstallOrUpdate()` was called only if not a connector before. So 1st commit seems safer.

Which version do you prefer?

Crash--:
I prefer the first one 'coz I find it easier to understand

Crash-- (approved):

PR #832, opened 2022-12-08
Fix : Keep underMaintenance active between search, reloads, filter change

cballevre:
```
### 🐛 Bug Fixes

* Show active categories with other filters than categories
* Keep underMaintenance active between search, reloads, filter change
```

### DIFF
--- a/src/ducks/apps/components/Sections/Sections.jsx
+++ b/src/ducks/apps/components/Sections/Sections.jsx
@@ -1,7 +1,6 @@
-import React, { useState, useEffect, useMemo } from 'react'
+import React, { useState, useEffect, useMemo, useRef } from 'react'
 import PropTypes from 'prop-types'
 import Fuse from 'fuse.js'
-import debounce from 'lodash/debounce'
 
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 import AppSections from 'cozy-ui/transpiled/react/AppSections'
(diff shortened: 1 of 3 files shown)


src/ducks/apps/components/Sections/Sections.jsx
```diff
@@ -27,53 +27,51 @@ import Filters from './components/Filters'
...
-
-  const onSearchChanged = useMemo(
-    () =>
-      debounce(value => {
-        const fuse = new Fuse(appsForSearch, options)
-        setSearchResults(value ? fuse.search(value) : null)
-      }, 300),
-    [appsForSearch, options]
-  )
+    const fuse = new Fuse(filteredApps, options)
```
Crash--, line 59: I know that this is not part of this PR, but this is weird, no? We should instanciate only once the `Fuse` object with all the apps, and then make a search on the filter and on the searchValue. We should not create this `Fuse` object every time, it doesn't make sense and should consume a lot of RAM.

Crash-- (approved):
I don't really like this solution, but let's move on.

src/ducks/apps/components/Sections/Sections.jsx
cballevre: Yes it's weird now you have point out. There is an other thing that disturb me, the filters are applied twice once before the search and once app list. In the same refacto, maybe we can use fuse.js for the two.

PR #835, opened 2022-12-12
feat: Rewording the consent page when installing a connector

Merkur39:
```
### ✨ Features

* Rewording the consent page when installing a connector
```

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -307,7 +307,7 @@
     "reason": "**Reason: **",
     "developer_default": "The developer of this application",
     "description": {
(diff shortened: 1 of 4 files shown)


src/locales/en.json
```diff
@@ -307,7 +307,7 @@
     "reason": "**Reason: **",
     "developer_default": "The developer of this application",
     "description": {
-      "internal": "This app will not transmit any data to its developer %{name}, it will use it only locally in your Cozy without transmission to anyone:",
+      "internal": "You are going to install a connector to facilitate the recovery of your data. It will access the following personal data but will never transmit them to anyone:",
```
Crash--, line 310: `%{name}` was used before, so can we remove it from the call?

PR #836, opened 2022-12-12
fix: Hide app under maintenance by default

cballevre:
```
### 🐛 Bug Fixes

* Hide app under maintenance by default
```

This PR is a fix for the feature to have an filter to show app under maintenance. I didn't apply a filter by default to hide app under maintenance without showing into query url params.

**Related PRs:**
- https://github.com/cozy/cozy-ui/pull/2312

### DIFF
--- a/src/ducks/apps/components/QuerystringSections.jsx
+++ b/src/ducks/apps/components/QuerystringSections.jsx
@@ -39,7 +39,12 @@ const parseURLSearchParams = urlParams => {
 
 const getFilterFromQuery = query => {
   const usp = new URLSearchParams(query)
-  return omit(parseURLSearchParams(usp), FILTER_BLACK_LIST)
+  const params = parseURLSearchParams(usp)
+  // Add showMaintenance by default to false
+  if (params.showMaintenance == undefined) {
+    params.showMaintenance = false
+  }
+  return omit(params, FILTER_BLACK_LIST)
 }
 
 const queryFromFilter = filter => {
@@ -66,7 +71,7 @@ class QuerystringSections extends React.Component {
     }
   }
 
-  componentWillUpdate(nextProps) {
+  UNSAFE_componentWillUpdate(nextProps) {
     if (this.props.location !== nextProps.location) {
       this.setState({
         filter: getFilterFromQuery(nextProps.location.search)
--- a/src/ducks/apps/components/Sections/components/Filters.jsx
+++ b/src/ducks/apps/components/Sections/components/Filters.jsx
@@ -10,18 +10,14 @@ import SettingIcon from 'cozy-ui/transpiled/react/Icons/Setting'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import useBreakpoints from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
-import { useEffect } from 'react'
 
 const { BarRight } = cozy.bar
 
 const Filters = ({ filter, onFilterChange }) => {
   const anchorRef = useRef()
   const [menuDisplayed, setMenuDisplayed] = useState(false)
-  const [appUnderMaintenance, setAppUnderMaintenance] = useState(true)
 
-  useEffect(() => {
-    setAppUnderMaintenance(filter.underMaintenance == undefined)
-  }, [filter.underMaintenance])
+  const areAppsUnderMaintenanceShown = filter.showMaintenance
 
   const toggleMenu = () => {
     setMenuDisplayed(!menuDisplayed)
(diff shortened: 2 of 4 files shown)


src/ducks/apps/components/Sections/components/Filters.jsx
```diff
@@ -32,17 +31,21 @@ const Filters = ({ filter, onFilterChange }) => {
...
   const hideAppUnderMaintenance = () => {
-    onFilterChange({
-      ...filter,
-      underMaintenance: false
-    })
+    if (appUnderMaintenance) {
+      delete filter.underMaintenance
+      onFilterChange({
+        ...filter
+      })
```
zatteo, line 47: Is it necessary to do
```
{
        ...filter
}
```

We can't pass just filter ?

src/ducks/apps/components/Sections/components/Filters.jsx
```diff
@@ -10,18 +10,17 @@ import SettingIcon from 'cozy-ui/transpiled/react/Icons/Setting'
...
   const [menuDisplayed, setMenuDisplayed] = useState(false)
-  const [appUnderMaintenance, setAppUnderMaintenance] = useState(true)
 
-  useEffect(() => {
-    setAppUnderMaintenance(filter.underMaintenance == undefined)
-  }, [filter.underMaintenance])
+  const appUnderMaintenance = useMemo(
+    () => filter.underMaintenance == undefined,
+    [filter.underMaintenance]
+  )
```
zatteo, line 23: What is the purpose of the useMemo ?

src/ducks/apps/components/QuerystringSections.jsx
```diff
@@ -39,7 +39,13 @@ const parseURLSearchParams = urlParams => {
...
 const getFilterFromQuery = query => {
   const usp = new URLSearchParams(query)
-  return omit(parseURLSearchParams(usp), FILTER_BLACK_LIST)
+  const params = parseURLSearchParams(usp)
+  if (params.underMaintenance == undefined) {
+    params['underMaintenance'] = false
+  } else {
+    delete params.underMaintenance
+  }
+  return omit(params, FILTER_BLACK_LIST)
```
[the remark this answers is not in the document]
Ldoppea: Also we should use either `params['underMaintenance']` form or `params.underMaintenance` form, but not both (this makes the code harder to read)

src/ducks/apps/components/Sections/components/Filters.jsx
```diff
@@ -10,18 +10,17 @@ import SettingIcon from 'cozy-ui/transpiled/react/Icons/Setting'
...
 
 const Filters = ({ filter, onFilterChange }) => {
   const anchorRef = useRef()
   const [menuDisplayed, setMenuDisplayed] = useState(false)
-  const [appUnderMaintenance, setAppUnderMaintenance] = useState(true)
 
-  useEffect(() => {
-    setAppUnderMaintenance(filter.underMaintenance == undefined)
-  }, [filter.underMaintenance])
+  const appUnderMaintenance = useMemo(
```
Ldoppea, line 20: nit: as we are editing this section, can we rename `appUnderMaintenance` variable with something like `areAppUnderMaintenanceShown` or `hasAppUnderMaintenanceFilter`?

With the current name my first understanding was that we were checking if an actual app was under maintenance, I didn't realize that it was about a filter.

src/ducks/apps/components/Sections/components/Filters.jsx
```diff
@@ -10,18 +10,17 @@ import SettingIcon from 'cozy-ui/transpiled/react/Icons/Setting'
...
+    () => filter.underMaintenance == undefined,
```
Ldoppea, line 21: Also this seems weird that `underMaintenance == undefined` means that `appUnderMaintenance = true`. I'm pretty sure renaming those variable would ease the comprehension

src/ducks/apps/components/QuerystringSections.jsx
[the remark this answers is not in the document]
cballevre: This makes it possible to hide applications by default without putting a filter in the url parameters. I don't know how to make it more explicit because the filters are managed by `cozy-ui` and I don't want to set an url parameter at the start. The other solution is refacto `cozy-ui` to repatriate app filtering in` cozy-store`.

zatteo (approved):

src/ducks/apps/components/Sections/components/Filters.jsx
```diff
@@ -10,18 +10,14 @@ import SettingIcon from 'cozy-ui/transpiled/react/Icons/Setting'
...
 
 const Filters = ({ filter, onFilterChange }) => {
   const anchorRef = useRef()
   const [menuDisplayed, setMenuDisplayed] = useState(false)
-  const [appUnderMaintenance, setAppUnderMaintenance] = useState(true)
 
-  useEffect(() => {
-    setAppUnderMaintenance(filter.underMaintenance == undefined)
-  }, [filter.underMaintenance])
+  const isAppUnderMaintenanceShown = filter.showMaintenance
```
Ldoppea, line 20: nit: as there are more that one app this may be `areAppsUnderMaintenanceShown`

Ldoppea (approved):
The `getFilterFromQuery` logic is way more simple to understand now 👍 

Don't forget to add the new cozy-ui PR in the related PRs list :) https://github.com/cozy/cozy-ui/pull/2312

PR #840, opened 2022-12-22
fix: Trash icon size was too small on uninstall button on some mobile screens

zatteo:
Happened on some mobile screens like 360*568. Solved by updating to new Buttons from cozy-ui.

```
### 🐛 Bug Fixes

* Trash icon size was too small on uninstall button on some mobile
```

### DIFF
--- a/src/ducks/apps/components/UninstallModal.jsx
+++ b/src/ducks/apps/components/UninstallModal.jsx
@@ -5,8 +5,9 @@ import compose from 'lodash/flowRight'
 
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
-import Button from 'cozy-ui/transpiled/react/Button'
-import Trash from 'cozy-ui/transpiled/react/Icons/Trash'
+import Buttons from 'cozy-ui/transpiled/react/Buttons'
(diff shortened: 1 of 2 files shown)


Ldoppea (approved):

src/ducks/apps/components/UninstallModal.jsx
```diff
@@ -5,8 +5,9 @@ import compose from 'lodash/flowRight'
 
 import { translate } from 'cozy-ui/transpiled/react/I18n'
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
-import Button from 'cozy-ui/transpiled/react/Button'
-import Trash from 'cozy-ui/transpiled/react/Icons/Trash'
+import Buttons from 'cozy-ui/transpiled/react/Buttons'
```
cballevre, line 8: You can still use `<Button>` by importing `import Button from 'cozy-ui/transpiled/react/Buttons'`

cballevre (approved):

src/ducks/apps/components/UninstallModal.jsx
zatteo: Is it the chosen way to use them ? Because I like to use Buttons if it Buttons I think it is more explicit.

PR #865, opened 2023-07-20
fix: Keep selected section and filters when navigating between Sections page and Application pages

Ldoppea:
When navigating from Sections page into Application page we lose the current selected category

This impact is that when we close the Application page using the `BACK TO THE LIST` button, then we cannot display the previously selected section and instead we display the full apps list

Another bug is that the previous search filter and the hide/show apps in maintenance filter are lost

This PR fixes this issue those bugs but also do some refactorings, and fix the Sidebar CSS so menu won't overflow anymore from the sidebar

### Before

When displaying the section page (top), the left menu overflows
The `type=konnector&category=insurance` is lost when navigating to the application page (bottom)
The Sidebar selection is lost when navigating to the application page (bottom)
image

### After

When displaying the section page (top), the left menu doesn't overflows anymore
The `type=konnector&category=insurance` is kept when navigating to the application page (bottom)
The Sidebar selection is kept when navigating to the application page (bottom)
image

Also the filter is kept when navigating back to the section page
image

```
### ✨ Features

*

### 🐛 Bug Fixes

* Selected category doesn't overflow anymore from the Sidebar
* Selected category is not lost anymore when selecting an App
* Search filter and other filters are not lost anymore when selecting an App

### 🔧 Tech

*
```

### DIFF
--- a/src/ducks/apps/components/ApplicationPage/Details.jsx
+++ b/src/ducks/apps/components/ApplicationPage/Details.jsx
@@ -34,21 +34,16 @@ export const Details = ({ app, description, changes, parent, mobileApps }) => {
   const { source, slug, langs, categories, developer, version } = app
 
   const appChannel = getChannel(source)
+  const isBeta = appChannel === REGISTRY_CHANNELS.BETA
+  const isDev = appChannel === REGISTRY_CHANNELS.DEV
 
   const client = useClient()
   const { t } = useI18n()
   const navigate = useNavigate()
   const { search } = useLocation()
+  const [displayBetaChannel, setDisplayBetaChannel] = useState(isBeta)
+  const [displayDevChannel, setDisplayDevChannel] = useState(isDev)
 
-  const [displayBetaChannel, setDisplayBetaChannel] = useState(
-    appChannel === REGISTRY_CHANNELS.BETA
-  )
-  const [displayDevChannel, setDisplayDevChannel] = useState(
-    appChannel === REGISTRY_CHANNELS.DEV
-  )
-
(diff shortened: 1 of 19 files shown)


src/ducks/apps/components/ApplicationPage/Details.jsx
```diff
@@ -30,203 +30,192 @@ const isValidUrl = url => {
...
+  const { source, slug, langs, categories, developer, version } = app
+
+  const appChannel = getChannel(source)
+
+  const client = useClient()
+  const { t } = useI18n()
+  const navigate = useNavigate()
+
+  const [displayBetaChannel, setDisplayBetaChannel] = useState(
+    appChannel === REGISTRY_CHANNELS.BETA
```
JF-Cozy, line 44: nit: on pourrait utiliser isBeta

test/ducks/apps/components/myApplications.spec.js
JF-Cozy, line 1: c'est plutôt un `test: blabla` et non un `fix: blabla` ce genre de commit je pense, il n'y a pas de modif en prod

test/ducks/apps/components/discover.spec.js
JF-Cozy, line 1: idem `test: ` et non `refactor: `

JF-Cozy (approved):

src/ducks/apps/components/ApplicationPage/Details.jsx
Ldoppea: Modifié ici: https://github.com/cozy/cozy-store/pull/865/commits/debc3e0d34ef6029801cbc6dfd4d8ad00651c527

PR #871, opened 2023-08-28
fix: Don't call cozy.bar outside of the React Tree

Crash--:
In a few scenarios, we don't inject the cozy.bar object to the react tree. This is the case for the Intent use case. Since we don't display the cozy-bar, we don't inject it in order to have a smaller bundle.

And within the code, if a component need the cozy.bar to be displayed and if it's called within an intent, then we polyfil the global var.

But since the destructuring object was done at the build / import phase and not during the component lifecycle, we got an error telling that we can't reader BarRight from undefined, even if the component was not displayed.

So to fix that, we should only call cozy.bar within the component lifecycle.


Merkur39 (approved):
Thanks 👍

PR #873, opened 2023-08-29
Fix(intent): Some content in ApplicationPage were not visible

Merkur39:
```
### 🐛 Bug Fixes

* In `INSTALL` Intent, some content in ApplicationPage were not visible
```

### DIFF
--- a/src/ducks/apps/components/ApplicationRouting/index.jsx
+++ b/src/ducks/apps/components/ApplicationRouting/index.jsx
@@ -21,8 +21,6 @@ const OutletWrapper = ({ Component }) => (
   </>
 )
 
-const intentStyle = { maxHeight: '100%' }
-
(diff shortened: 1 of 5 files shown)


src/ducks/apps/components/ApplicationRouting/index.jsx
```diff
@@ -45,8 +43,9 @@ export class ApplicationRouting extends Component {
 
     return (
       <div
-        className="sto-modal-page"
-        style={intentData ? intentStyle : undefined}
+        className={`sto-modal-page${
+          intentData ? ' sto-modal-page-intent' : ''
+        }`}
```
Crash--, line 48: nit: cx() can be useful in that case ;)

Crash-- (approved):

PR #886, opened 2024-01-22
fix: Reduce click area on app vote component (VO-172)

acezard:
Decided not to update cozy-ui for this issue.
Handling it locally with already existing css classes


cballevre (approved):
This solution will correct the problem, but we'll need to review the implementation for this part, as there's a problem in the cozy-store, since the cozy-ui component has no style.

PR #911, opened 2024-06-27
[SCR-683] feat: Add new categories to the store and their locales

KillianCourvoisier:
```
### ✨ Features

* Adds "clouds", "finance" and "tech" in known categories
* Adds new locales for each new categories

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "cozy-minilog": "^3.3.1",
     "cozy-realtime": "3.14.4",
     "cozy-stack-client": "^48.2.0",
-    "cozy-ui": "^110.1.0",
+    "cozy-ui": "^110.2.0",
     "emoji-js": "3.7.0",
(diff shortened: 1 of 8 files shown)


cballevre:
Thanks for your PR but it will miss the part about cozy-ui. If you want to have section title you should add your translation here also : https://github.com/cozy/cozy-ui/blob/master/react/AppSections/locales/en.json pls 🙏

KillianCourvoisier:
Sure, didn't noticed that 🙏

KillianCourvoisier:
Hop @cballevre https://github.com/cozy/cozy-ui/pull/2663 here you are, can you do the review here et there if it all ok for you ?

cballevre (approved):
This PR looks good to me 👍 You just need to update cozy-ui with the new version before merge it

cballevre:
@acezard are you aware of this new translation for the home ? Maybe should we add them

PR #913, opened 2024-07-03
feat: display custom apps in store

acezard:
Tests will fail until cozy-ui is updated

image


src/ducks/apps/components/Discover.jsx
```diff
@@ -55,6 +56,7 @@ export class Discover extends Component {
                 onAppClick={this.onAppClick}
                 intentData={intentData}
                 parent="discover"
+                alternativeStore={alternativeStore}
```
cballevre, line 59: Why for `Discover` you concat apps and alternativeApps and here you share alternativeStore directly?

src/ducks/apps/components/MyApplications.jsx
```diff
@@ -68,10 +70,13 @@ const MyApplicationsWrapper = props => {
   const isExact = useMatch('myapps')
   const navigate = useNavigateNoUpdates()
   const [searchParams] = useSearchParams()
+  const { alternativeApps } = useAlternativeStore()
+  const installedAltApps = alternativeApps.filter(app => app.installed)
```
cballevre, line 74: Maybe we could add this filter into the useAlternativeStore so we would be able to reuse it in other place

src/ducks/AlternativeStore/tests/transformData.spec.ts
cballevre, line 1: nit: More little case would be more digest because show all the different cases you are testing

src/ducks/AlternativeStore/fixtures/flag.json
cballevre, line 1: I would prefer random naming in test like the one we discuss in daily

src/ducks/AlternativeStore/transformData.ts
```diff
@@ -0,0 +1,57 @@
...
+    config.fileTypeMappings
+  )) {
+    fileTypeMappings[fileType] = categoryName
+  }
+
+  return data.map(file => {
+    const metadataType = file.metadata.type || 'default' // Handle undefined type
+    const filePath = file.path
+
+    if (!filePath) return file
```
cballevre, line 26: By default file don't have a path. From what I understood, you keep all shortcut even if they are not in a folder from the config 🤔

src/ducks/apps/components/SidebarCategories.jsx
```diff
@@ -85,6 +90,8 @@ export const SidebarCategories = ({
   const linksArray = options.map(cat => {
     if (cat.value === 'konnectors') {
       return renderLink(cat, `${parent}?type=konnector${extraParams}`, location)
+    } else if (cat.value === 'shortcuts') {
+      return renderLink(cat, `${parent}?type=file${extraParams}`, location)
```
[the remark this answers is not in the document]
acezard: Well technically the type is file so I thought it would be simpler here to use the type (like konnectors and apps) instead of the class

test/ducks/apps/__snapshots__/index.spec.js.snap
```diff
@@ -181,56 +181,8 @@ Array [
...
-          "3.0.0",
-        ],
-        "stable": Array [
-          "3.0.0",
-        ],
-      },
-    },
-    "lang": "en",
-    "type": "FETCH_APP_SUCCESS",
+    "error": [Error: Error while getting the application with slug: collect],
```
cballevre, line 184: I think there is an error inside the test 🤔

src/ducks/apps/components/SidebarCategories.jsx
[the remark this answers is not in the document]
cballevre: Let's go with that, even if it makes less sense from the user's point of view

src/ducks/apps/components/SidebarCategories.jsx
```diff
@@ -61,10 +64,12 @@ export const SidebarCategories = ({
   let appsList = []
   switch (parent) {
     case '/discover':
-      appsList = apps
+      appsList = apps.concat(alternativeApps)
       break
     case '/myapps':
-      appsList = installedApps
+      appsList = installedApps.concat(
+        alternativeApps.filter(app => app.installed)
```
cballevre, line 71: You can reuse `installedAlternativeApps` here also :)

src/ducks/apps/components/ApplicationPage/Details.jsx
```diff
@@ -48,12 +48,14 @@ export const Details = ({ app, description, changes, parent, mobileApps }) => {
     !!categories.length &&
     categories.map(c => t(`app_categories.${c}`))
   const developerName =
-    developer && getTranslatedManifestProperty(app, 'developer.name', t)
+    (developer && getTranslatedManifestProperty(app, 'developer.name', t)) ||
```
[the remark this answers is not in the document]
cballevre: I think it's still complex to read, but we'll leave it at that for now 😕

src/ducks/apps/components/ApplicationPage/index.jsx
```diff
@@ -109,11 +109,9 @@ export class ApplicationPage extends Component {
       'short_description',
       t
     )
-    const appLongDesc = getTranslatedManifestProperty(
-      app,
-      'long_description',
-      t
-    )
+    const appLongDesc =
```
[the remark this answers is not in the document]
cballevre: Same as above

src/ducks/apps/components/SidebarCategories.jsx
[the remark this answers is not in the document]
cballevre: From a user point of view, I think it makes less sense, but let's leave it at that 👍

test/ducks/apps/__snapshots__/index.spec.js.snap
acezard: cozy-stack-client had to be updated to work with cozy-client. I didn't notice that when updating cozy-client

src/ducks/apps/components/SidebarCategories.jsx
```diff
@@ -68,7 +68,7 @@ export const SidebarCategories = ({
       break
     case '/myapps':
       appsList = installedApps.concat(
-        alternativeApps.filter(app => app.installed)
```
cballevre: I think there's a blunder here 🤔 This shouldn't be 
```

appsList = installedApps.concat(installedAlternativeApps)
```

cballevre (commented):
The most important thing are done just last few kilometres to the finish line 🦾

src/ducks/AlternativeStore/useAlternativeStore.ts
```diff
@@ -0,0 +1,39 @@
...
+import { useQuery } from 'cozy-client'
+import { IOCozyFile } from 'cozy-client/types/types'
+import flag from 'cozy-flags'
+import { useExtendI18n } from 'cozy-ui/transpiled/react/providers/I18n'
+
+export const useAlternativeStore = (): {
+  alternativeApps: AlternativeShortcut[]
+  installedAlternativeApps: AlternativeShortcut[]
+} => {
+  const shortcutsQuery = buildShortcutsQuery()
```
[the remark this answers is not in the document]
acezard: @cballevre do you have an idea how we should handle this scenario?

src/ducks/apps/components/ApplicationPage/Details.jsx
```diff
@@ -48,12 +52,14 @@ export const Details = ({ app, description, changes, parent, mobileApps }) => {
...
+    (isShortcutFile(app) && app.attributes?.metadata?.source)
   const developerUrl =
-    developer && getTranslatedManifestProperty(app, 'developer.url', t)
+    (developer && getTranslatedManifestProperty(app, 'developer.url', t)) ||
+    app.attributes?.metadata?.url
   const shortVersion = version && version.match(/^(\d+\.\d+\.\d+)-.*$/)
   const displayedVersion =
-    (shortVersion && shortVersion.length && shortVersion[1]) || version
+    ((shortVersion && shortVersion.length && shortVersion[1]) || version) ?? '-' // If no version, display a dash
```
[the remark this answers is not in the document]
acezard: just to be sure to understand, this is related to the updated snapshots right? as in, they shouldn't have been updated since the tests don't have shortcut data

src/ducks/AlternativeStore/useAlternativeStore.ts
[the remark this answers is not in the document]
cballevre: I think I see 2 options : 
- (best one) revert the way to search shortcut by searching folder first by path in the flag and after search for shortcut by dir_id with a `useQueryAll` to ensure having all files
- use the current version and transform the `useQuery` into `useQueryAll` but not optimal at all but it can be a first step

PR #933, opened 2024-11-18
[PRD-538] Hide apps by flag

paultranvan:
```
### ✨ Features

* Hide apps by flag `apps.hidden`

### 🐛 Bug Fixes

*

### 🔧 Tech

*
```

### DIFF
--- a/test/ducks/apps/__snapshots__/index.spec.js.snap
+++ b/test/ducks/apps/__snapshots__/index.spec.js.snap
@@ -19,6 +19,54 @@ Array [
     "apps": Array [
       Object {
         "categories": Array [
+          "cozy",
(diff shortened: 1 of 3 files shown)


test/ducks/apps/__snapshots__/index.spec.js.snap
```diff
@@ -17,6 +17,54 @@ Array [
   },
   Object {
     "apps": Array [
+      Object {
```
cballevre, line 20: nit: Perhaps we can remove the test that generates this case because we needed to check that collect wasn't displayed

cballevre (approved):
LGTM 👍

